PR #358, opened 2012-06-12
Set Encoding.default_external to UTF-8 to build manual on systems with non Unicode locale

boutil:
Hi!

When trying to build the manual on a system with non Unicode locale (i.e. C) with Ruby 1.9, the compilation of the manual fails because Ruby has to read strings supposed to be ASCII but containing Unicode characters.
Adding the single line:

```
Encoding.default_external = "UTF-8" if defined? Encoding
```

at the top of manual/manual/manual.rb fixes this issue.

Thanks.

### DIFF
--- a/manual/manual/manual.rb
+++ b/manual/manual/manual.rb
@@ -2,6 +2,9 @@
 #
 # Generates the Prawn by example manual.
 #
+
+Encoding.default_external = "UTF-8" if defined? Encoding
+
 require File.expand_path(File.join(File.dirname(__FILE__),
                                    %w[.. example_helper]))
 


dookim:
where is  manual/manual/manual.rb?? in window ?

sandal:
@doomkim: This issue is ancient! If you're looking for the entry point into our manual, you can find it here: https://github.com/prawnpdf/prawn/blob/master/manual/contents.rb

The manual building code has been extracted into its own gem, its repo can be found here: https://github.com/prawnpdf/prawn-manual_builder

All `rake manual` does is run the above contents.rb file. This is the task in case you're interested though:

```ruby
desc "Generate the 'Prawn by Example' manual"
task :manual do
  puts "Building manual..."
  require File.expand_path(File.join(File.dirname(__FILE__),
    %w[manual contents]))
  puts "The Prawn manual is available at manual.pdf. Happy Prawning!"
end
```

Hope that helps!

Issue #409, opened 2012-10-08
Table, rotated text in a cell bug


ottost:
Hi all,

I'm having problems with rotated text, use this code and the result you can see it in the picture. How I can do to make the text fill the height of the cell?

You can see the problem in the image http://cl.ly/K1IC

The code is:

data = Array.new
data << [{content: 'Estructuras', height: 100, valign: :bottom, align: :center, rotate: 90}]
pdf.table(data)

pdf.move_down(15)

data = Array.new
data << [{content: 'Estructuras', height: 100, valign: :bottom, align: :center, rotate: 90, width: 20}]
pdf.table(data)

Thanks in advance.

Issue #112, opened 2010-05-09
Bad cell size when using :rotate inside a table 


nicolas-mosconi:
I'm trying to create a table with vertical headers:

Prawn::Document.generate("header.pdf") do
  header = %w[Name Occupation]
  table([header]) do
    row(0).style(:style => :bold, :rotate => 90)
  end
end

The cell size doesnt fit the content size.

Sorry for my bad english

PR #505, opened 2013-06-17
Better table cell rotation

straydogstudio:
Here is some initial work to improve table cell rotation. Right now Prawn doesn't stay within the cell (partly because the text is wrapped before the content is rotated, and partly because rotate_around moves text outside the cell for all but the simplest of cases.) This fix keeps the text wrapped correctly on rotations of 9, 90, 180, 270, etc. Intermediate angles are at better than they were for now. Here is what it looks like:

prawn_table_rotate

Note that only table height is specified on that table. 

I've created Table::Cell::Box and Table::Cell::Formatted::Box which inherit from Prawn::Text::Formatted::Box. There is also Table::Cell::Formatted::Wrap which inherits from  Prawn::Core::Text::Formatted::Wrap. Only a few methods are overridden. If there is a better way to organize this, please let me know.

Intermediate angles still don't fit in the box, but they wrap reasonably well. The swap of height/width for wrapping happens on 45 degree angles. If this approach makes sense, sometime I'll tackle the wrap logic for other angles, and the text can still fit on an angle. 

One last thing. This code ignores rotate_around. I overrode valid_options so including rotate_around throws an error. Perhaps it should just be ignored?

Needed: 
- Vertical / horizontal alignment
- Wrapping in intermediate angles?
- Rotation may need to be adjusted for padding changes. 

Added a rotate test at the end of table_spec.rb. cell_spec.rb adjusted for new classes. All tests are passing. 

Other: I added .ruby-version and .ruby-gemset to .gitignore for the latest [rvm](https://github.com/wayneeseguin/rvm/issues/1694)

Issues referenced: [409](https://github.com/prawnpdf/prawn/issues/409#issuecomment-18233381) & [112](https://github.com/prawnpdf/prawn/issues/112)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 nbproject
 pkg 
 .rvmrc
+.ruby-version
+.ruby-gemset
 .bundle
 Gemfile.lock
 drop_to_console.rb
--- a/lib/prawn/table.rb
+++ b/lib/prawn/table.rb
@@ -13,6 +13,9 @@ require 'prawn/table/cell/text'
 require 'prawn/table/cell/subtable'
 require 'prawn/table/cell/image'
 require 'prawn/table/cell/span_dummy'
+require 'prawn/table/cell/formatted/wrap'
+require 'prawn/table/cell/formatted/box'
+require 'prawn/table/cell/box'
 
 module Prawn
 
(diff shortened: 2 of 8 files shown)


hmt:
Is this still active? I need a working solution for 90 degrees text in my table headers. Would really appreciate any solution. Thank you for your work!

practicingruby:
@hmt: See this: https://github.com/prawnpdf/prawn-table/pull/32  -- the main blockers right now are documentation and maybe some code cleanup, but you can try running from @straydogstudio's fork for the moment... it should work!

hmt:
@practicingruby oh, I don't know how I could miss that PR. I looked in all the wrong places.

Issue #496, opened 2013-05-25
Bug: Less-than sign with `inline_format`


jlduran:
``` ruby
text "'<' is the <i>less-than</i> sign"
move_down 25
text "'<' is the <i>less-than</i> sign", :inline_format => true                                
```

Produces:

> '<' is the <i>less-than</i> sign
> 
> '' is the _less-than_ sign

Issue #527, opened 2013-08-08
:inline_format => true Bug


nithinstany:
pdf.text "< <b> 3000</b>", :inline_format => true -  its only print 3000 is in bold "<" is missing. Is there any solution

PR #565, opened 2013-10-13
Make less than sign work with inline format

gettalong:
Due to the inline formatting of text, using `<` did not work.

This pull request (includes specs and updated API documentation) makes using `<` without escaping work. This is done by using a more complex regex string that takes the recognized HTML-esque tags into account.

Fixes issues #496, #527


johnnyshields:
Think we should revisit this logic. Prawn supports only a subset of full HTML so this sort of thing makes a lot of sense to me.

johnnyshields:
I've raised PR #889 to expose `escape` and `unescape`

Issue #779, opened 2014-10-05
Improve awareness of lack of UTF-8 support in PDF built-in (AFM) fonts


practicingruby:
Even though Prawn supports UTF-8 just fine when working with TTF font files, the default behavior is to use PDF built in fonts (e.g. Helvetica.afm). 

These fonts are the handful of special fonts that are expected to be provided by the PDF viewer, preventing the need to embed them in documents. However, they are _not_ UTF-8 compatible, and are in an obscure encoding format (Win-1252, which is similar but not identical to Latin-1).

We make an effort to transcode UTF-8 strings into Win-1252 when built-in AFM fonts are used, but in practice this is not suitable for internationalized text, because so few glyphs are supported. This understandably causes confusion for users, and we need to find a way to help them.

Possibly solutions include:
- A prominent note in the manual, both in the introduction section and the text sections.
- A warning or error raised whenever built-in fonts are used with text that has unsupported 
  characters in it.
- (More radical) A warning displayed whenever an AFM font is used, that can be globally disabled with, `Prawn::Font.disable_afm_warning` flag.
- (Most radical) Ship Prawn with a handful of TTF fonts and discontinue support for AFM entirely.

The last two options are probably not realistic, but I'm considering them only because we have a ton of tickets that pile up due to these issues.

@prawnpdf/contributors @prawnpdf/maintainers Thoughts?

Issue #777, opened 2014-10-01
Encoding mismatches in text boxes with truncation


airblade:
Here's my use case: I have a fixed size text box and I want to write text in it.  Often that text is too long for the text box and so the text should be truncated.  This all works fine for me.

However I would like to indicate when the text has been truncated, perhaps with an ellipsis or simply three full stops (periods).  This is how I am trying to achieve it:

```ruby
box = Prawn::Text::Box.new text, disable_wrap_by_character: true, ...  # sizing options omitted
overflow = box.render dry_run: true
text = text.sub(/ #{overflow}$/, '...') unless overflow.empty?
document.text_box text, disable_wrap_by_character: true, ...  # as above
```

However I often, though not always, get `incompatible encoding regexp match (ASCII-8BIT regexp with UTF-8 string)` errors.  Sometimes `text` is UTF-8, sometimes it's US-ASCII.  Sometimes `overflow` is UTF-8, sometimes it's US-ASCII.

I have tried re-encoding both `text` and `overflow` to UTF-8 but then the substitution doesn't work because the strings no longer match.

- Is this the best approach for achieving my goal?
- If it is, how can I make it work? :)

PR #793, opened 2014-10-19
Change text box to return remaining text as UTF-8, improve Win1252 handling internally, raise errors or warnings rather than silently replacing invalid glyphs

practicingruby:
Currently if using an AFM-font,Prawn's text box will return overflow text in Win1252 encoding, even though the user must provide their text in UTF-8 format. To render this overflowed text, the user needs to pass `:skip_encoding`, which is a fairly awkward workflow.

In an ideal setting, we should treat any transcoding to Win1252 (<strike>which isn't even supported by Ruby, so we implement it ourselves</strike>) as an implementation detail, and never allow the transcoded text to cross the public API boundary. In other words, Prawn should be UTF-8 in, UTF-8 out to keep things simple for users.

See related discussion on #777, and on #779.

### DIFF
--- a/lib/prawn/document.rb
+++ b/lib/prawn/document.rb
@@ -66,7 +66,7 @@ module Prawn
 
     VALID_OPTIONS = [:page_size, :page_layout, :margin, :left_margin,
                      :right_margin, :top_margin, :bottom_margin, :skip_page_creation,
-                     :compress, :skip_encoding, :background, :info,
+                     :compress, :background, :info,
                      :text_formatter, :print_scaling]
 
     # Any module added to this array will be included into instances of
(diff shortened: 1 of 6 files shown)


isaiah:
> a68eef6 contains a scary but probably good change: It raises an error if non-Windows-1252 glyphs are attempted to be rendered in AFM fonts. The behavior of existing versions of Prawn is to replace with an underscore, which is pretty much never what users will want, and masks bugs.

@practicingruby I think this change makes the fallback_fonts ignored.
I have a document that contains Cyrillic, CJK, Arabic, Greek or anything, so I cannot really tell the document which font to use, previously I depends on the fallback_fonts feature, the UTF-8 glyphs are rendered correctly, but it's broken with 2.0.0

practicingruby:
@isaiah Do not mix AFM fonts with fallback fonts... use TTF fonts throughout and you'll be fine!

isaiah:
@practicingruby How can I do that? Currently I monkey patched the Prawn::Document#font_families method and #fallback_fonts method, to add my own TTF fonts. Should I disable the original fonts completely?

PR #819, opened 2015-02-03
Example of a #document method added to the view.rb manual page

giuseb:
A couple of explanatory sentences added too.


giuseb:
Thanks!

I actually debated whether to mention the `@document` approach. The reason I did not is that it feels a bit like breaking encapsulation, as in--the user of `Prawn::View` needs to know the name of the specific instance variable, and the maintainer of the library is not free to change it later on. Whereas a method feels more like a deliberate API choice.

But I am not a sophisticated Ruby programmer, so maybe perusal of an instance variable defined in a mixin is a perfectly good practice...

sandal:
@giuseb Every use of a module is a shot-gun blast to encapsulation: There's no major difference between overwriting methods or variables, because they're both in a single shared space. 

However, you should take a look at how Prawn::View is implemented... 

```ruby
    def document
      @document ||= Prawn::Document.new
    end
```

As long as we always only refer to `document()` within the mixin and the only reference to `@document` is in this method, we're not going to run into issues. We will only use your `@document` variable if you don't define your own `document()` method.

sandal:
To expand on this a little further:

Suppose you have code like this:

```ruby
class Foo
  include Prawn::View

   def initialize
     @fancy_custom_document = ....
   end

   def document
     @fancy_custom_document
  end
end
```

Then the following things are true:

* `Prawn::View` will use your `document()` method.
* `Prawn::View` will not create a `@document` variable.
* `Prawn::View` will ignore a `@document` variable if you create one.
* Everything will work fine, as long as `@fancy_custom_document` points to a Document-like thing.

So we're not leaking any state-based implementation details, unless you explicitly opt-in to it.

giuseb:
Yes, everything you write makes perfect sense, but still...

Are there many other places in Prawn where the user needs to get somewhat cozy with the underlying implementation?

How about something like this:

```ruby
module Prawn
  module View
    def document
      @document ||= Prawn::Document.new(@doc_options || {})
    end

    def document_options(opts)
      @doc_options = opts
    end

    ...

  end
end

class Shrimp
  include Prawn::View
  def initialize(opts={})
    document_options opts
  end
end

Shrimp.new(page_size: 'A4')
```

Where you are not telling people *about* `Prawn::View`. Rather, you are basically saying, if you need to customize the document, call `document_options` before invoking `Prawn::Document` methods.

People are still free to read the source and decide that they'd rather override `#document` or define their own `@document`, but from the perspective of the concise API presented in the beautiful manual, something along these lines would seem to me fairly straightforward. Existing code should not be affected but such an addition, right?

Makes sense?

sandal:
I've thought of doing something like that, but that actually introduces more state and a wider interface, giving even more potential for collision. I'm happy with the design as it is.

giuseb:
Okay, fair enough. Thanks for all the good work, Greg!

Giuseppe

dsisnero:
I like this approach of delegating to an internal pdf document.  However, in asciidoctor-pdf the document method is already taken by the internal asciidoctor document.  It would be better if either the name could be supplied or if it could be changed to something not likely to be used. __pdf_document.

practicingruby:
@dsisnero: Consider changing the name in asciidoctor-pdf instead? We're upstream.

mojavelinux:
@practicingruby is absolutely right. This is not Prawn's conflict to resolve. Asciidoctor PDF needs to be refactored as a Prawn View (see https://github.com/asciidoctor/asciidoctor-pdf/issues/53). When that is done, we can rename the reference to the Asciidoctor document object as needed. Further comment should be directed to the linked issue.

pointlessone:
@giuseb Could you please rebase this PR on master?

pointlessone:
@giuseb The pages don't look good. Could you please take a look?

page-1
page-2

giuseb:
My commits DO change the manual, thereby failing the recently added spec (see commit 1444f4). Please advise!

pointlessone:
This a new addition we've been working for a while. Manual reviews historically have been a huge time sink during releases. But not every release actually had manual changes in them. This failure lets us know that we need to review the manual.

Make sure the manual looks as expected and change the hashes in the failing spec. Also I personally would appreciate if you post screenshots of the changed pages.

manual/graphics/helper.rb
```diff
@@ -20,6 +20,6 @@
   )
   stroke_axis(
     at: [140, 140], width: 200, height: cursor.to_i - 140,
-    step_length: 20, negative_axes_length: 40, color: 'FF00'
+    step_length: 20, negative_axes_length: 40, color: 'FF0000'
```
pointlessone: This is fixed on master: 0712aeb8370043ebe2ef2c6273be799ad4d1d5e8

giuseb:
Please let me know if the manual page looks good now. If so, I will update the PR, ok?

screen shot 2017-02-21 at 10 35 19

pointlessone:
Looks good.

giuseb:
I see that travis fails because the jruby-generated hash for the manual is missing.
Unfortunately, I don't have jruby on my machine (can't seem to get the properly JDK installed on macOS 10.12...)

Any chances someone else might compute that hash?

pointlessone:
I'll handle that.

pointlessone:
@giuseb I've rebased and merged this PR. I'll close this PR now. Feel free to reopen.

NidalGhonaim:
I thought i was crazy. I just started learning prawn today and i couldn't find the "document method" documented in the manual "basic_concepts/view.rb" ... I had to google so many things with nothing relavent until i just started randomly opening issues until i found this one X'D ... i'm both happy and want to cry ..

PR #822, opened 2015-02-06
Fix styled width_of when using TTFs.

ernie:
Previously, width_of calculations on styled text were relying on the
document font's `name` attribute in order to look up the appropriate
font style. This doesn't work for TTF fonts, since the name is a full
path to a single style of font, and the Prawn must know about the font
family in order to find another style.

This update uses the family instead, which allows the width_of
calculations to work properly with TTFs, and doesn't appear to cause any
regressions based on existing specs.

I know very little about fonts, to be honest. This PR was borne out of a difficulty being encountered in production code that uses TTFs. So, it's entirely possible I'm missing some edge case with other font types. Maybe DFONTs, for instance, don't require Prawn font_families to be updated, since they're all in the file?


packetmonkey:
@sandal I bundled this branch as the version of Prawn used by my print on demand site and it didn't break any of my tests either.

sandal:
This looks like a good change. It may take me some time to get around to reviewing, but as long as I don't find any major issues we'll try to get this merged in time for the next release.

(Which is hopefully soon, but truthfully has been delayed for a while)

ernie:
Awesome! Glad I could help out!

practicingruby:
See #827.

Issue #797, opened 2014-10-27
`Prawn::Document#on_page_create` in v 1.3.0


BrentWheeldon:
Hi there,

We have some code that does, effectively:

```
document = Prawn::Document.new(...)
# stuff
document.on_page_create do |doc|
 # more stuff
end
```

Since upgrading to 1.3.0 we get: `NoMethodError: undefined method 'on_page_create' for #<Prawn::Document:0x007f92f8a88fd0>` when running this code. I checked the changelog, but didn't see anything about this method, and I can't see it in the source anymore. I can see it that it's been removed from `lib/prawn/document/internals.rb` in b48b20603f823a390f1be2cfd8d5aa1995a5e7c5, but it's not clear if there's an alternative?

Any help would be greatly appreciated!

nathansamson:
~~~
document.on_page_create do |doc|
 # more stuff
end
~~~

Obviously only works for the second page and later, unless you do pass `skip_page_creation: true` on document creation...

So even if on_page_create is resurected it feels a bit clunky... Repeaters with different z-indexes sounds nice though

sandal:
@nathansamson: Yeah, in time for the next release I'm only going to restore the behavior because that's how it's been since basically forever, but that is a good wart to think about dealing with when we produce a better alternative in the future.

jessedoyle:
@practicingruby - I submitted PR #825 that delegates the method call to `renderer`. Hopefully we can get this in the 2.0.0 release!

practicingruby:
Merged!

bousquet:
:+1: Thanks.  Used this today in combination with prawn-svg for crisp background image underneath flowing text.

PR #825, opened 2015-02-19
Delegate on_page_create method to renderer. #797

jessedoyle:
A simple patch that delegates `on_page_create` calls to `renderer.on_page_create`.


practicingruby:
Looks good! Now that we have a CHANGELOG.md file in the repository, please try to update it in future pull requests. I can always revise or expand upon notes, but this ensures that we won't entirely miss documenting a change.

PR #827, opened 2015-02-20
Fix styled width_of when using TTFs. (Revised version of #822)

practicingruby:
From @ernie:

> Previously, width_of calculations on styled text were relying on the
document font's `name` attribute in order to look up the appropriate
font style. This doesn't work for TTF fonts, since the name is a full
path to a single style of font, and the Prawn must know about the font
family in order to find another style.

> This update uses the family instead, which allows the width_of
calculations to work properly with TTFs, and doesn't appear to cause any
regressions based on existing specs.


practicingruby:
@ernie:

I added a test to make sure that bold differed from plain DejaVuSans to help make sure we're not getting a false positive test result. I also removed what I believe to be an old and unnecessary default value (I can't think of why we'd want to use "Helvetica" if the font doesn't exist. It's our default font, but the font object should be present and the family can be read from it.)

Please test this against your production code and let me know if we're ready to :shipit:.

ernie:
Thanks! And sorry, didn't realize it was safe to rely on font.family being available based on the previous code!

practicingruby:
@ernie: I'm not 100% certain it is safe, but the manual and tests pass, and if this breaks something I'd rather fix the root cause than rely on this weird default.

ernie:
@practicingruby Also, yep, this looks to do the trick. Thanks again!

practicingruby:
Squashed and merged in https://github.com/prawnpdf/prawn/commit/13cd1c427f7ecd775b8244d6395dc814334ea8c3 (there is a little bit of merge noise there, but it should be harmless)

practicingruby:
@ernie: Because your pull request was accepted, I've given you commit access to all of Prawn's repositories. Please read our [contribution guidelines](https://github.com/prawnpdf/prawn/wiki/Contributor-welcome-notes), and thanks for the patch!

PR #834, opened 2015-02-27
manual/text/positioned_text.rb: Improve wording

paulmenzel:
Fix up commit c709a25f (Added positioned text examples.) by replacing
*Given that said, …* by *Given that, …*.

*That said, …* would have been fine too.


packetmonkey:
Looks good, thanks for the tweak!

PR #844, opened 2015-04-29
Reproducible build of the manual

boutil:
Hi,

The following two commits are contributions for the Debian reproducible builds effort
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=782890
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=782891
in order to allow reproducible builds of the Debian packages, by removing every bit that can change in a non predictable way. In particular, they allow the manual to be built (bitwise) identically accross builds, and probably also PDF generated by prawn more generally.
The contributions are:
- use deterministic keys (integers increasing by one) for gradients instead of #hash
- use BUILD_EPOCH environment variable instead of Time.now to timestamp the manual, if available.
  Thanks for considering including these changes.

Cédric

### DIFF
--- a/lib/prawn/graphics/patterns.rb
+++ b/lib/prawn/graphics/patterns.rb
@@ -60,7 +60,7 @@ module Prawn
       end
 
       def gradient_registry_key(gradient)
-        if gradient[1].is_a?(Array) # axial
+        key = if gradient[1].is_a?(Array) # axial
           [
             map_to_absolute(gradient[0]),
             map_to_absolute(gradient[1]),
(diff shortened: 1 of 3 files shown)


packetmonkey:
Hey Cédric, thanks for the fixes. I would like to see a couple changes though before I'm comfortable pulling this in.

1. Can you add an entry to the changelog for these changes and include links to the original debian bugs.
2. Can you include tests for these changes so we don't break them in future updates and know why that code is needed
3. I'm not super happy with the [logic figuring out the last_update for the cover](https://github.com/prawnpdf/prawn/blob/reproducible_build/manual/cover.rb#L26-L34). I think I would rather like to see something more obviously defaulting to the git repo latest commit, then falling back to Time.now allowing it to be overwritten by the BUILD_EPOCH environment variable.

Maybe something like

```ruby
if Dir.exist?("#{Prawn::BASEDIR}/.git")
  commit = `git show --pretty=%h`
  git_commit = "git commit: #{commit.lines.first}"
  last_update = Time.at(`git log -1 --pretty='%ct'`.to_i)
end

unless last_update
    git_commit = ""
    last_update ||= Time.now
end

if ENV['BUILD_EPOCH']
  last_update = Time.at ENV['BUILD_EPOCH'].to_i
end
```

That way the cascade of options is more obvious.

Thoughts?

pointlessone:
I believe this has been fixed a while ago when we added a spec for manual changes.

I'll close this PR now. Feel free to reopen if the issue is still present.

Issue #801, opened 2014-11-07
Links inside a stamped text are not clickable


hidakatsuya:
My environments are:
* MacOSX 10.10
* Prawn 0.12.0, 1.3.0
* Ruby 1.9.3, 2.1.3

```ruby
Prawn::Document.generate 'foo.pdf' do
  create_stamp 'bar' do
    text '<b>Prawn</b> <link href="http://github.com">GitHub</link>', inline_format: true
  end
  stamp 'bar'
end
```

Formats except link are properly rendered, but links are not clickable.

Issue #498, opened 2013-06-01
Links inside a repeater not working


jlduran:
I don't think this should be the normal behavior, but:

```ruby
repeat :all do
  text "<link href='http://example.com'>external link</link>", inline_format: true
end
```

produces a PDF with no link.

The `/URI (http://example.com)` is there, however, does not work.

sandal:
I've confirmed that #801 is a direct reproduction of this problem at the stamp level, so please follow that ticket for future updates.

PR #845, opened 2015-04-29
Fix links/annotations within stamps. #801

jessedoyle:
This is a fix for a known bug that has been present within Prawn for some time now. Previously, links were not clickable in stamps because annotations were only referred from the stamp's XObject. This patch adds the necessary annotation references to the page object that the stamp was called from.

This patch should also be a fix for links within repeaters, as mentioned in #498.

_(Sorry, it looks like my text editor removed trailing spaces from the changelog - hopefully this doesn't clutter up the commit)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 ## PrawnPDF master branch
 
-No new changes yet!
+### Links in repeaters/stamps are now clickable
+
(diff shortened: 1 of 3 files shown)


packetmonkey:
This looks good to me and fixes the issue. I'll merge it with the stripped change log.

Thanks Jesse!

PR #846, opened 2015-04-30
Add new release policy to README.md

packetmonkey:
With the merging of #845 we now have new code in master that hasn't been released. I'm proposing we commit to a regular 6 week release to roll up and release any new commits to help prevent development stagnation. I proposed 6 weeks because that is the release cycle used by other major open source projects (Google Chrome, Firefox, Ember.js) and I have found it easy to keep up to date with updates to those projects when I needed to. 

If no code has been merged into master during a 6 week cycle, we wouldn't cut a release. I would expect most releases to be a minor version bump. If we have a major feature release we can discuss making a minor (or major) version bump and any backwards incompatible changes would also be a major release, in line with the current versioning policy.

An open question is do we release a version if the only changes that are merged are internal refactoring or other non-behavior changing commits. 

At this point I am personally committing to cut the prawn releases per whatever we decide.

I'm interested in anyone's feedback but especially looking for input from @cheba 

If this is adopted I would plan to cut the first release 6 weeks after that adoption, unless we decide otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 Prawn is a pure Ruby PDF generation library that provides a lot of great functionality while trying to remain simple and reasonably performant. Here are some of the important features we provide:
 
 * Vector drawing support, including lines, polygons, curves, ellipses, etc.
-* Extensive text rendering support, including flowing text and limited inline formatting options. 
+* Extensive text rendering support, including flowing text and limited inline formatting options.
 * Support for both PDF builtin fonts as well as embedded TrueType fonts
 * A variety of low level tools for basic layout needs, including a simple grid system
 * PNG and JPG image embedding, with flexible scaling options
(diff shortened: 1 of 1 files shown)


jessedoyle:
I'm a big fan of a stable, predictable release cycle. Furthermore, I think that it is worthwhile to release a new version even if it's a standard code refactor. Simply refactoring code can lead to performance improvements (possibly), plus I know that I've hacked the sources in my ruby gems profile before - better readability is always a good thing!

packetmonkey:
I'm going to leave this open until next wednesday to give anyone else an opportunity to discuss this. As of right now I plan on merging this PR then and beginning the 6 week release cycle.

Thanks for the :+1: @jessedoyle

packetmonkey:
I'm going to go ahead and run with this plan. I'll plan to cut the next release in 6 weeks which currently is around June 19th.

PR #864, opened 2015-05-05
Enforce consistent multi-line string concatenation

packetmonkey:
This PR enforces consistent multi-line string concatenation prefering to use `\` instead of the previously used `>>` and `+`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,9 +14,12 @@ Encoding:
 AsciiComments:
   Enabled: false
 
-# Due to some layout constraints in our examples we want to allow this rule to
+# Due to some layout constraints in our examples we want to allow these rule to
 # be ignored in the manual.
 Style/ClosingParenthesisIndentation:
+  Exclude:
+    - manual/**/*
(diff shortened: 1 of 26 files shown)


crazymykl:
A lot of these would read better as HEREDOCs.

packetmonkey:
Long term I expect a lot of them to go away. 

They seem to break down into 3 main groups

- 95% Long test names
- ~10 Long exception messages
- ~2 Long strings related to formatted text parsing

I expect to refactor the spec suite so we use more context blocks and fewer really-long named tests.

The text parsing in lib/prawn/text/formatted/line_wrap.rb and lib/prawn/text/formatted/parser.rb I'll probably leave alone until a possible future where we add better support for pluggable text formatters. That leaves roughly 10 the long exception strings which do fall more on the side of documentation and may be worth looking into.

packetmonkey:
I'm going to go ahead and merge this in to keep making progress on the general consistency/cleanup.  Feel free to open an issue or PR to talk about where we could use heredocs to clean things up further!

Issue #807, opened 2014-12-12
lib/prawn/text.rb - consider deleting '#' characters 


ghost:
Hi,

Not sure if sandal considers this good or not - if not it can just be closed. Thanks.

Currently you can do this code:

```
text 'Ruby in Red', :color => 'FF0000'
text 'Ruby in Red', :color => '#FF0000'
```

The first variant works fine, and displays the text in red.

The second line displays the code in green (???)

Screenshot:
![alt text](http://i.imgur.com/beq9Z7D.png)

My proposals:
(a) Leading '#' characters will be ignored (this behaviour should be documented). The effect would be that the second line would be treated like the first line (unless perhaps there was a reason for this which I overlooked, but to me right now this seems more like a tiny bug)
(b) The green colour here, I am not sure if it makes sense. I assume that currently the '#' marks this as invalid? In that event, perhaps it should default to black? (e. g. black text on white background, by default)

practicingruby:
@packetmonkey: Do you still want to work on this? Since it's a small patch, I could probably write it myself if you don't plan to do it in the next few days, so that we can include it in the 2.0.0 release.

packetmonkey:
This totally dropped off my radar, I'll try to knock it out this weekend but I won't feel hurt if you beat me to it.

practicingruby:
I'll drop a note on this ticket if I start working on this, otherwise go ahead and give it a shot.

packetmonkey:
Going to assign the issue to myself so it shows up on my GH dashboard.

practicingruby:
I realized that this isn't a release blocker because we'd just be turning an undefined behavior into an exception. So rather than rushing to build a fix before the 2.0.0 release (which may happen today or tomorrow), just fix this when you get around to it @packetmonkey.

shevegen:
Gregory, it is time to say it - I <3 your work on prawn. :)  (There was an older "port", ruby-freepdf (fpdf) loosely based on the php-pdf, but it never was high quality code, very difficult to navigate and modify; prawn is much much better from a technical point of view)

practicingruby:
Thanks @shevegen. Prawn is very much the work of dozens of contributors. Most of the good ideas aren't mine. :-P

petergoldstein:
@pointlessone This issue should be closed, as a behavioral decision was made and merged as discussed in #869

PR #869, opened 2015-05-05
Raise error for colors with leading #

tomprats:
Saw this [issue](https://github.com/prawnpdf/prawn/issues/807) and thought I could help! Really appreciate all the work you guys do!

### DIFF
--- a/lib/prawn/graphics/color.rb
+++ b/lib/prawn/graphics/color.rb
@@ -94,7 +94,11 @@ module Prawn
       def color_type(color)
         case color
         when String
-          :RGB
+          if color =~ /\A[0-F]{6}\z/i
+            :RGB
+          else
+            fail ArgumentError, "Unknown type of color: #{color.inspect}"
+          end
         when Array
           case color.length
           when 3
--- a/manual/graphics/color.rb
+++ b/manual/graphics/color.rb
@@ -10,7 +10,7 @@ filename = File.basename(__FILE__).gsub('.rb', '.pdf')
 Prawn::ManualBuilder::Example.generate(filename) do
   stroke_axis
 
-  # Fill with Yellow using RGB
+  # Fill with Yellow using RGB (Unlike css, there is no leading #)
   fill_color "FFFFCC"
   fill_polygon [50, 150], [150, 200], [250, 150],
                [250, 50], [150, 0], [50, 50]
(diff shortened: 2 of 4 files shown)


packetmonkey:
Nice! Thanks for the PR!

First off please add an entry to the change log so we can doc the updated behavior.

Also I'm not sure that checking for the length of the string will be enough, as someone could say pass `EEFFGG`. What do you think about updating the logic to
```ruby
when String
  if color =~ /\A[0-F]{6}\z/i
    :RGB
  else
    raise ArgumentError, "Unknown type of color: #{color.inspect}"
  end
```

Thanks again!

tomprats:
Good call! I wasn't thinking about the actual string content, I'll update both of those

tomprats:
I'm just going to close this because it's probably no longer valid

packetmonkey:
Hey I missed that you updated this branch, if you are still interested in getting it in I'm open to merging.

Thanks!

tomprats:
@packetmonkey Thanks! Should be good to go now

manual/graphics/color.rb
```diff
@@ -10,7 +10,7 @@
 Prawn::ManualBuilder::Example.generate(filename) do
   stroke_axis
 
-  # Fill with Yellow using RGB
+  # Fill with Yellow using RGB (Unlink css, there is no leading #)
```
crazymykl: Unlike.

lib/prawn/graphics/color.rb
```diff
@@ -94,7 +94,11 @@ def process_color(*color)
       def color_type(color)
         case color
         when String
-          :RGB
+          if color =~ /\A[0-F]{6}\z/i
```
pointlessone: There are a few characters that we probably don't want to match between `'9'` and `'A'`. I'd prefer this regex to be something more specific. For example `/\A[0-9a-f]{6}\z/i`.

pointlessone:
@tomprats If you're steal interested, please address my comment and rebase your branch on master. Thank you.

pointlessone:
@tomprats Than you for your contribution. I've rebased this PR on master and merged. GitHub didn't pick it up so I will close it now.

tomprats:
@pointlessone Thanks! Sorry I never got around to making the extra changes

PR #879, opened 2015-05-19
Add pixel conversion

klebervirgilio:
Hi,

I'm working in the generation of  PDF's. Nothing too complex, I've gotten a few html templates then I need to sure the pdf's looks as similar as possible the templates.

A several facts don't allow me to use gems like  wickedpdf,  pdfkit, imgkit.

So, I did miss the `px2pt` conversion method, was wondering why it didn't exist? I think I'm not the first looking for method.  

Anyway I thought would be good to ask you guys with a PR.

http://endmemo.com/sconvert/pixelpoint.php
http://stackoverflow.com/questions/139655/convert-pixels-to-points

Thanks!

### DIFF
--- a/lib/prawn/measurement_extensions.rb
+++ b/lib/prawn/measurement_extensions.rb
@@ -47,4 +47,8 @@ class Numeric
   def pt
     return pt2pt(self)
   end
+
+  def px
+    return px2pt(self)
+  end
 end
--- a/lib/prawn/measurement_extensions.rb
+++ b/lib/prawn/measurement_extensions.rb
@@ -48,7 +48,7 @@ class Numeric
     return pt2pt(self)
   end
 
-  def px
-    return px2pt(self)
+  def css_px
+    return csspx2pt(self)
   end
 end
--- a/lib/prawn/measurements.rb
+++ b/lib/prawn/measurements.rb
@@ -66,5 +66,13 @@ module Prawn
     def pt2mm(pt)
       return pt * 1 / mm2pt(1) # (25.4 / 72)
     end
+
+    def px2pt(px)
+      return px * 0.75
+    end
+
+    def pt2px(pt)
+      return pt / 1.333333
+    end
   end
 end
--- a/lib/prawn/measurements.rb
+++ b/lib/prawn/measurements.rb
@@ -67,11 +67,11 @@ module Prawn
       return pt * 1 / mm2pt(1) # (25.4 / 72)
     end
 
-    def px2pt(px)
+    def csspx2pt(px)
       return px * 0.75
     end
 
-    def pt2px(pt)
+    def pt2csspx(pt)
       return pt / 1.333333
     end
   end
--- a/spec/measurement_units_spec.rb
+++ b/spec/measurement_units_spec.rb
@@ -18,5 +18,6 @@ describe "Measurement units" do
     1.ft.should == 72 * 12
     1.yd.should == 72 * 12 * 3
     1.pt.should == 1
+    1.px.should == 0.75
   end
 end
--- a/spec/measurement_units_spec.rb
+++ b/spec/measurement_units_spec.rb
@@ -18,6 +18,6 @@ describe "Measurement units" do
     1.ft.should == 72 * 12
     1.yd.should == 72 * 12 * 3
     1.pt.should == 1
-    1.px.should == 0.75
+    1.css_px.should == 0.75
   end
 end


cheba:
It doesn't exist because there's no universal points to pixels conversion. It very much depends on the pixel density of the target medium.

Your code works for CSS pixels. But those are not real pixels. It would be confusing to introduce a unit that has little relevance both to its name and its real counterpart outside of CSS.

klebervirgilio:
Oh! I see, makes sense! Thanks @cheba.

mojavelinux:
I need this method too. Perhaps we can compromise and call it it csspx2pt. While there is no univeral points to pixels conversion, then there's the real world where this works most of the time. So this is useful.

klebervirgilio:
@mojavelinux that was what I did on my end. I guess it might help others!

mojavelinux:
Cool. We'll definitely be using this in Asciidoctor PDF, whether it makes it into Prawn or not.

klebervirgilio:
Sounds good! How Can I help? Let me know if you guys need anything else.

mojavelinux:
> How Can I help?

Are you asking about Prawn or Asciidoctor PDF?

klebervirgilio:
To be honest, till now I didn't even know about Asciidoctor PDF. I guess the best path would be to look at the issues?

mojavelinux:
> I guess the best path would be to look at the issues?

Indeed. Many of them relate to Prawn and how we use it. I'm fixing a lot of the little stuff right now.

The big (but fun and challenging) issue is how we are organizing the code. I didn't know Prawn very well when I started it, so we ended up with a monolithic subclass of Document. We really need to upgrade Prawn and migrate to a Prawn View. The first step is a simple demonstration of how that might look. I have no problem porting the code if we have a clear idea of where we are going. See https://github.com/asciidoctor/asciidoctor-pdf/issues/53

The other major issue is inline and floated images. I have an idea about how to do it, but I wish it were something that Prawn could handle natively. See https://github.com/asciidoctor/asciidoctor-pdf/issues/9

Prawn is critical to the future of Asciidoctor PDF because, honestly, there is really no other library or tool that produces half-way decent PDFs...nothing that comes close to Prawn. The only other real contender (for our stack) is iText. But we need changes to the Asciidoctor and AsciidoctorJ APIs before we can start that experiment...so we continue to push forward with Prawn.

mojavelinux:
And to clarify, aside from being a converter from AsciiDoc to PDF, the goal of Asciidoctor PDF is to provide the fastest way to create a professional-looking PDF document from scratch. This is particularly value for writers, but I use it for all my business documents (contracts, agreements) as well.

echo -e "= Document Title\nAuthor Name\n\nCreating a PDF doesn't get much easier than this." | asciidoctor-pdf -o sample.pdf -

packetmonkey:
I agree with @cheba that it doesn't feel right to be in Prawn. One solution may be to use ruby's open classes to define the method yourself and use it such as

```ruby
module Prawn
  module Measurements
    def px2pt(px)
      return px * 0.75
    end

    def pt2px(pt)
      return pt / 1.333333
    end
  end
end
```

I'm also open to a possible API extension to allow people to add their own measurements for ease of use. This could enable the creation of a prawn-pixelpoint gem (or whatever) that anyone who needs the functionality can require in, which I believe would help your use case @mojavelinux in addition to letting people support whatever they need (my next project is now going to use furlongs just so I can say it in meetings). But I'm not sure what benefit it would provide beyond option 1 other than possible name collisions but I think the developer should be aware of that anyway.

I think option 1 works, can be packaged as a gem and depended on, and it keeps us from becoming a giant warehouse of every possible conversion for every use case.

Full working example

```ruby
require 'prawn'

module Prawn
  module Measurements
    def furlong2pt(furlongs)
      furlongs * 7_920.004 * 72
    end
  end
end


Prawn::Document.generate 'output.pdf' do
  extend Prawn::Measurements

  x       = in2pt 3
  y       = cm2pt 7.620
  width   = yd2pt 0.083
  height  = furlong2pt 0.000379

  stroke do
    rectangle [x, y], width, height
  end
end
```

Thoughts? I'll reopen the PR while we come up with a solution that works for everyone.

mojavelinux:
> One solution may be to use ruby's open classes to define the method yourself and use it such as

That's the approach I was planning to take. If you are using Ruby 2+, you also have prepend as a way to do it.

> my next project is now going to use furlongs just so I can say it in meetings

Solid. Bring back the furlong!

> I'll reopen the PR while we come up with a solution that works for everyone.

I think we should either add the methods `css_px2pt` and `pt2ccs_px` or do nothing. I don't really see an API extension point here as these are just utility classes.

What is useful is to know about this relationship. It took me over a year to discover it. Having something in the manual would have been a huge help to me in the beginning...even if you say in the manual that it isn't a universal thing, but likely will map in a predictable way in most cases...that would have helped me think about the transition from web to PDF a heck of a lot better. So much stuff makes sense now that made no sense before.

mojavelinux:
According to the CSS3 spec <http://www.w3.org/TR/css3-values/>, “1px is equal to 1/96th of 1in” and “1pt is equal to 1/72th of 1in”, which is how you get the conversion of 1px = 0.75pt.

klebervirgilio:
IMHO, I would say Ruby by itself offers a good API to add dynamic features/methods/... I mean, I don't see a good reason to create an API. 
@mojavelinux nice one! :+1: bringing some facts to the conversation!

mojavelinux:
@klebervirgilio It's also funny that the CSS spec basically says that pixels is a magical unit. It really only means anything for print output, in which case the 1/96th of an inch is required. On a screen, a pixel is "a visible single line", whatever that actually means :)

cheba:
"a visible single line" is still not quite a physical pixel on the screen. Ref: [A Pixel Identity Crisis](http://alistapart.com/article/a-pixel-identity-crisis)

packetmonkey:
It sounds like everyone's use case is handled by adding methods to Prawn::Measurements themselves and if it's super useful feel free to create a gem and let the mailing list know to claim your unit conversion helper method glory!

If someone wants to open a new PR updating the documentation to help make this point clearer, I would be happy to check it out.

Thanks for the discussion everyone, re-closing the PR.

PR #889, opened 2015-06-08
Expose escape and unescape methods in Text::Formatted::Parser module

johnnyshields:
This PR exposes the `escape` and `unescape` logic for Prawn HTML. Currently this logic is inline buried inside other methods.

### DIFF
--- a/lib/prawn/text/formatted/parser.rb
+++ b/lib/prawn/text/formatted/parser.rb
@@ -30,6 +30,14 @@ module Prawn
           Regexp.new(regex_string, Regexp::MULTILINE)
         end
 
+        ESCAPE_CHARS = {
+          '&' => '&amp;',
+          '>' => '&gt;',
+          '<' => '&lt;'
+        }.freeze
+
+        UNESCAPE_CHARS = ESCAPE_CHARS.invert.freeze
+
         def self.format(string, *_args)
           tokens = string.gsub(%r{<br\s*/?>}, "\n").scan(PARSER_REGEX)
           array_from_tokens(tokens)
@@ -94,8 +102,7 @@ module Prawn
               suffix = '</color>'
             end
 
-            string = hash[:text].gsub('&', '&amp;').gsub('>', '&gt;')
-              .gsub('<', '&lt;')
+            string = escape(hash[:text])
             prefix + string + suffix
           end.join
         end
@@ -221,8 +228,7 @@ module Prawn
                 /character_spacing='([^']*)'/.match(token)
               character_spacings << matches[1].to_f unless matches.nil?
             else
-              string = token.gsub('&lt;', '<').gsub('&gt;', '>')
-                .gsub('&amp;', '&')
+              string = unescape(token)
               array << {
                 text: string,
                 styles: styles.dup,
@@ -237,6 +243,14 @@ module Prawn
             end
           end
           array
+        end
+
+        def self.escape(text)
+          text.gsub(Regexp.union(ESCAPE_CHARS.keys), ESCAPE_CHARS)
+        end
+
+        def self.unescape(text)
+          text.gsub(Regexp.union(UNESCAPE_CHARS.keys), UNESCAPE_CHARS)
         end
       end
     end
--- a/spec/prawn/text/formatted/parser_spec.rb
+++ b/spec/prawn/text/formatted/parser_spec.rb
@@ -664,4 +664,22 @@ describe Prawn::Text::Formatted::Parser do
       expect(described_class.array_paragraphs(array)).to eq(target)
     end
   end
+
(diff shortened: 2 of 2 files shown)


pointlessone:
@johnnyshields If you're still interested in getting this merge could you please rebase this PR on master?

johnnyshields:
Rebase + tests passed, please merge

lib/prawn/text/formatted/parser.rb
```diff
@@ -238,6 +236,14 @@ def self.array_from_tokens(tokens)
           end
           array
         end
+
+        def self.escape(text)
+          text.gsub('&', '&amp;').gsub('>', '&gt;').gsub('<', '&lt;')
```
pointlessone: Since, we are refactoring this, maybe we can use a replacement has and spare use a few extra passes over the string?

lib/prawn/text/formatted/parser.rb
```diff
@@ -238,6 +236,14 @@ def self.array_from_tokens(tokens)
...
+        end
+
+        def self.unescape(text)
+          text.gsub('&lt;', '<').gsub('&gt;', '>').gsub('&amp;', '&')
```
pointlessone: Additionally, we may move the mapping elsewhere to make sure that `escape` and `unescape` perform mirror transformations.

lib/prawn/text/formatted/parser.rb
johnnyshields: What do you mean here? Like `text.gsub(/&|>|</) {|val| if val == '7' then ... }` I think that's not significantly faster and is more complex.

lib/prawn/text/formatted/parser.rb
johnnyshields: OK where shall I move it? To a class constant?

johnnyshields:
I think the two comments would unnecessarily increase complexity and propose to merge PR as-is. We can do those refactors in a separate PR.

lib/prawn/text/formatted/parser.rb
pointlessone: Yes, constant will do.

lib/prawn/text/formatted/parser.rb
pointlessone: `gsub` can take a hash as the second argument.

```ruby
text.gsub(/&|>|</, '&' => '&amp;', '>', => '&gt;', '<' => '&lt;')
```

This whole hash  can live in a constant. Let's call it `MAPPING` for example.

Then we can reduce it further.

```ruby
MAPPING = {
  '&' => '&amp;',
  '>', => '&gt;',
  '<' => '&lt;',
}.freeze

def self.escape(text)
  text.gsub(Regexp.union(MAPPING.keys), MAPPING)
end

def self.unescape(text)
  mapping = MAPPING.inverse
  text.gsub(Regexp.union(mapping.keys), mapping)
end
```

johnnyshields: cool! was not aware that gsub can take a hash as an arg. learned something new.

pointlessone:
Could you please clean up history a bit and rebase on master?

johnnyshields:
@pointlessone why not just squash commits with GH?

pointlessone:
The resulting commit message is ugly.

johnnyshields:
@pointlessone please merge

pointlessone:

aried3r:
Hey! I just ran into this issue again by mistake and found this PR. Is this still something that could make it into the code base? I read through https://github.com/prawnpdf/prawn/pull/565, https://github.com/prawnpdf/prawn/issues/527 and https://github.com/prawnpdf/prawn/issues/496 and it seems that this PR was deemed and acceptable solution to the problem.

We'll probably use this and parts of https://github.com/prawnpdf/prawn/pull/565 in our code base for now, but an official solution would be very welcome. :)

petergoldstein:
@pointlessone or @johnnyshields  Can you please rebase this?  Assuming spec pass (including Rubocop) it looks like it should just be merged.

pointlessone:
For whatever reason GitHub doesn't give me an option to rebase this branch. 🤔

aried3r:
> For whatever reason GitHub doesn't give me an option to rebase this branch. 🤔

My guess would be that this PR is older than the option to let maintainers push to a PR branch, see [this doc](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

petergoldstein:
Ok, I'll build an updated PR with the cherrypicked commit.

petergoldstein:
@pointlessone Now that #1252 is merged this can be closed.

Issue #891, opened 2015-07-05
Gradients don't work when user space has been transformed


mogest:
I'm the author of prawn-svg, an SVG to PDF renderer that uses prawn, and have had a few requests to implement SVG gradients.  @mojavelinux is interested in this for his asciidoctor-pdf project.

Unfortunately, prawn doesn't implement gradients quite right.  PDF gradients/patterns take coordinates in the coordinate space of the document, not the "user space", so if you perform a scale/rotate/translate and then paint a gradient inside, it doesn't render correctly.

I have a working solution to this, but it's not super elegant, so I thought I'd bring it up here for discussion before I tidy it enough to be presentable.

The idea is that all calls to <tt>transformation_matrix</tt> result in the matrix being pushed on to a document-scope stack, and <tt>save_graphics_state</tt> and <tt>restore_graphics_state</tt> do what you'd expect them to do with the stack before passing the call on to PDF::Core.

The gradient code has then been extended to matrix multiply the transformations on the stack together, and pass to the PDF <tt>Matrix</tt> attribute in the shading pattern, instead of using an identity matrix as it currently does.

It works, but is a substantial change.  It also means we'd be able to implement a working version of <tt>skew</tt>, I believe.

Any objections to the approach before I submit a PR?

packetmonkey:
Thanks for the info @mogest, to be honest I'm not 100% comfortable with that code or that part of the spec so I'll take your word at that it's wrong :)

My first question is would implementing your change break backwards compatibility? If so we would need to implement your fix in a way that we can enable the new correct behavior while preserving the old behavior until we release Prawn 3.0 and remove the deprecated wrong code.

I would say throw open a PR so we can look over what you have and see what can be done. I'll try to become more knowledgeable about the area as well so we can come up with the right solution.

cheba:
@mogest Is my understanding correct that the problem is here: https://github.com/prawnpdf/prawn/blob/master/lib/prawn/graphics/patterns.rb#L65-L66 (and the relevant calls in the other branch)?

mogest:
@cheba, it's actually the Matrix that's passed on https://github.com/prawnpdf/prawn/blob/master/lib/prawn/graphics/patterns.rb#L124 that's causing the problem.  The matrix has to match the current transformations that have been effected in the document.

I've created a pull request @packetmonkey.  I'd love to hear what everyone thinks, as I feel like it might be a bit of a contentious solution.

packetmonkey:
[...]

PR #894, opened 2015-07-19
Track transformations and apply to gradients' matrices

mogest:
PDF gradients/patterns take coordinates in the coordinate space of the
document, not the "user space", so if you perform a scale/rotate/translate
and then paint a gradient inside, it doesn't render correctly.

This commit tracks transformations applied to the document, and multiplies
the gradient matrix with this tracked transformation matrix so that the
gradient appears in the correct place in the document.

This has been created in response to issue #891.


mogest:
Rubocop wants <tt>.map(&:to_f)</tt> instead of <tt>.map { |n| n.to_f }</tt>, but I was just following the style of the surrounding code.  Happy to change.

cheba:
@mogest Please do.

Code seems clean enough but I haven't got to actually try it out yet. Hopefully will have some time this week.

packetmonkey:
My main concern with this, beyond trying to find the time to truly understand this change, is any backwards compatibility implications.

From my basic understanding, it looks like it a user could be relying on the broken behavior of their gradient's not being rotated, and this change would alter their behavior if they upgraded.

Am I understanding correctly?

mojavelinux:
>  is any backwards compatibility implications.

If that's a major concern, why not just add an option at the top-level to use legacy or compliant gradient logic?

One thing is for certain. We have to get past this or else gradients in SVGs embedded in PDFs (which is a major use case for a PDF generator) is never going to work correctly.

packetmonkey:
I was thinking something similar. For that to work we would have to let users opt into the new correct behavior, deprecate the old broken behavior, then when we release prawn 3 at some point in the future we can remove the old code and the logic for handling it's optional use and just use the correct new behavior. 

I don't believe Prawn has done something like that before so I want to make sure that a breaking change is transitioned correctly, there are a lot of people possibly using the bad behavior and not knowing it.

The other option is that merging this PR necessitates the next version of prawn released is 3.0, but I'm not personally comfortable releasing a breaking 3.0 right now so soon after 2.0.

I would love ideas and any discussion on the ramifications of those choices.

mojavelinux:
> For that to work we would have to let users opt into the new correct behavior, deprecate the old broken behavior, then when we release prawn 3 at some point in the future we can remove the old code and the logic for handling it's optional use and just use the correct new behavior.

Precisely how I envisioned it.

>  there are a lot of people possibly using the bad behavior and not knowing it.

Do you really think there are that many people using gradients directly? I've been using Prawn for almost 2 years and I never saw a need for it (the need comes in the form of the infrastructure necessary for libraries like prawn-svg to work).

I tend to agree that it is too soon for Prawn 3, so the best path forward (if the agreement is that this is a critical breaking change) is to allow it to be opt-in. prawn-svg could even cause this switch to be enabled when activated (that's up to prawn-svg).

cheba:
I use both prawn-svg and gradients in my documents. But I don't use transformations.

I don't think changing core behaviour in prawn-svg is a nice thing to do. Probably, we can create a new gem for this very specific thing. Warn in Prawn when gradients used with transformations that it's going to change in 3.0 and that there's this one gem you should use to get correct behaviour. prawn-svg can warn users too when incorrect behaviour is evoked (but prawn would probably detect it anyway).

mogest:
I feel like the external gem might be a bit fiddly given the change touches three separate classes.  I'd suggest that we add an option to the fill_gradient/stroke_gradient call, something like <tt>apply_transformations: true</tt>.  It'll initially default to false, and maybe log a warning if you don't explicitly set it false or true.

As you suggest @packetmonkey, it'd make sense to change the default to true in prawn 3.

Downside to the approach is the <tt>apply_transformations</tt> option hangs around afterwards.  It's a pretty simple switch, so I don't think it'll negatively impact code complexity to keep it there.

mojavelinux:
I like that proposal. (I'm not sure where the impression came from of
adding a separate gem. I wouldn't want it to be solved that way either.)

packetmonkey:
I think this sounds like the best plan, lets make the new behavior opt-in for now and when we cut prawn v3 we will make it the default behavior and remove the old.

This also introduces a new question, how do we mark it as deprecated. I'm inclined to throw a `Kernel#warn` call in the legacy code path with a link to a wiki page describing the change and what they should do to enable the new behavior. Normally I would also mark the method as deprecated in the docs but we aren't changing that, so we may want to add a note there as well in case they look at the docs.

Thoughts?

cheba:
+1 on `throw`. +1 on wiki. +1 on docs.

mogest:
OK, <tt>:apply_transformations</tt> option added, <tt>warn</tt> displayed if option not explicitly supplied and a transformation was made (no point warning the user if turning on the option wouldn't result in different behaviour, IMO.)  Added some info to the method comments, and a link to a non-existent wiki page!

packetmonkey:
Alright I took some time to check out the branch and play with it. Everything is looking good in my tests when I used a scale block and a rotate block. 

I think this is working as good as it can be.

I went ahead and set up the [wiki page](https://github.com/prawnpdf/prawn/wiki/Gradient-Transformations). I'm open to any feedback / edits on making that easier for developers to understand as this is the first time we are really executing a deprecation migration plan in Prawn, so if there is a precedent we want to set, now is the time.

I'm looking for another :+1: or two but I think this is ready to go in, for sure before the next prawn release.

@cheba Agree? 

@mogest Just throwing it out there that you rock :) Thanks a _ton_ for pulling this together.

packetmonkey:
@mogest Also please add an entry to the CHANGELOG for this.

mojavelinux:
Great job on the wiki page @packetmonkey and another big thumbs up to @mogest for implementing the new behavior. Great teamwork!

cheba:
:star:

mogest:
Thanks everyone!  And thanks for the great contributing experience.

Wiki page looks great, changelog committed, ready to roll hopefully!

packetmonkey:
Aaannddddd merged :)

All glory to prawn-svg 

giphy


Thanks again!

PR #897, opened 2015-07-30
Refactor arranger

packetmonkey:
While hunting down a performance problem I did a little work cleaning up the specs and some of the text arranger. They are mostly minor and straightforward and I added a few missing specs I saw as I went.

### DIFF
--- a/lib/prawn/text/formatted/arranger.rb
+++ b/lib/prawn/text/formatted/arranger.rb
@@ -31,7 +31,7 @@ module Prawn
         end
 
         def space_count
-          unless @finalized
+          unless finalized
             fail "Lines must be finalized before calling #space_count"
           end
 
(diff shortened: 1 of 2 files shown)


lib/prawn/text/formatted/arranger.rb
```diff
@@ -212,6 +219,8 @@ def font_style(styles)
 
         private
 
+        attr_accessor :finalized
```
cheba: This seem to be deprecated in Ruby 2. Doesn't it warn you?

`protected` seem to be fine for `attr_accessor`, though.

packetmonkey: Matz agreed that this warning is wrong and it's going to be removed in a future version.

https://bugs.ruby-lang.org/issues/10967

cheba: Ah, cool!

PR #905, opened 2015-09-02
Update document.rb

dsisnero:
add ability to pad top and bottom with one method.  Short circuit move up , down if n == 0.  This comes from asciidoctor-pdf

### DIFF
--- a/lib/prawn/document.rb
+++ b/lib/prawn/document.rb
@@ -419,6 +419,7 @@ module Prawn
     # the current bounding box.
     #
     def move_up(n)
+      return if n == 0
       self.y += n
     end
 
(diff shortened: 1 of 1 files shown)


bvogel:
Shouldn't this be tested and documented in the manual as well?

packetmonkey:
I think I would prefer to see a new method that called the existing methods twice instead of adding complexity to the existing methods

Maybe something like

```ruby
def pad_vertical(top_padding, bottom_padding)
   move_down top_padding
   yield
   move_down bottom_padding
end
```

Which could possibly be combined with an `#indent` call to give you a padded box somewhere.

We would also for sure need tests for this, and maybe include it in the manual.

Thoughts?

pointlessone:
@dsisnero Thank you for your contribution.

I don't feel like this is a useful addition. It is simple enough to keep it in your document rendering code if you need it.

I will close this PR now. Feel free to reopen if you have anything to add to the discussion.

PR #910, opened 2015-09-18
Add Blend Modes

iamjohnford:
First, let me apologize since I don't know enough about the PDF spec, and structure of Prawn, this code probably isn't in the right place.

It looks like the last time someone mentioned blend modes was in 2010 (https://groups.google.com/forum/#!searchin/prawn-ruby/blend$20mode/prawn-ruby/h0Htg6YV_vE/Vg6SqTOqvrkJ). I needed to overlay images with the Multiply blend mode so I learned just enough of the PDF spec and Prawn to add it to Transparency.

Here's one way to test the blend modes. It works best if you add a background to the page.

```ruby
Prawn::Graphics::Transparency::BLEND_MODES.each_with_index do |blend_mode, index|
  transparent(1, 1, blend_mode) do
    fill_color [100, 0, 0, 0]
    fill_rectangle [0, bounds.height - (index * 50)], bounds.width, 50
  end
end
```

The spec mentions that PDFs can accept an array of blend modes but this code only adds one. I wasn't sure how to add an array of values to the PDF. If you think blend modes would be a useful addition, and it should be written in another way, I'd be happy to restructure the code if you can help point me in the right direction.

Thanks!


packetmonkey:
@iamjohnford Thanks for the pr!

@mogest @cheba Could I get your input on this? I'm unfamiliar with this part of the spec as well and am currently not in a position to triage this with regard to how correct this solution is.

@iamjohnford Can you also add a spec for this new functionality? I'm up for getting the feature merged in the right way and we need to make sure it will continue to work in the future. It would also be useful to have an example of how to use this code so we can easily see it in action and work though it.

mogest:
Code looks good @iamjohnford.  Checks out with the PDF spec too.

Two thoughts:

First, I feel like it might be nicer to have a separate `blend_mode` method, instead of overloading the `transparent` method. If you want to set the transparency and blend mode, you could just stack them as gs operator calls are cumulative. Just a bit cleaner so you can use blend mode by itself and not change the transparency setting.  What do you think?

    transparent 0.5 do
      blend_mode :Multiply
        ... do stuff ...
      end
    end

Second, might as well support the array syntax.  You just supply an array of symbols to `ref!` instead of just the one symbol, and it'll take care of it for you.

Specs would be good, as well as an example in the manual and an entry in the changelog!

Yell out if you'd like a hand, I'm happy to help.

iamjohnford:
Thanks! Those tips helped point in the right direction. I've started creating a separate BlendMode module. 

Should it pretty much mimic the Transparency code? I have some working code where I basically duplicated the Transparency module and changed the appropriate items to BlendMode. 

The arrays also work (at least I see it when viewing the PDF source) but I haven't found an example that will create a different effect with a combination of two blend modes. I'm still trying to figure out if that's correct.

mogest:
Yeah, it's gonna be pretty much the same as Transparency.  Most prawn methods wrap the call in `save_graphics_state` and `restore_graphics_state` and yield if they're given a block, and just make the call without saving/restoring if they're not given a block; it might be nice for your code to follow that standard, even though Transparency only handles the block-provided case (from memory.)

Re multiple blend modes, they're not cumulative; here's 7.5.2 in the PDF spec:

> Its value is either a name object, designating one of the standard blend modes listed in Tables 7.2 and 7.3 on pages 520 and 524, or an array of such names. In the latter case, the application should use the first blend mode in the array that it recognizes (or Normal if it recognizes none of them). Therefore, new blend modes can be introduced in the future, and applications that do not recognize them have reasonable fallback behavior. (See implementation note 72 in Appendix H.)

packetmonkey:
:+1: for the separate blend_mode method, and we will want a spec specifically to cover making sure multiple blend modes are applied in the correct order to allow the reader to fallback as 7.5.2 noted.

Thanks so much!

iamjohnford:
Okay, I need some feedback on the current commit. Here are some things that were added:

* Updated the changelog
* Added BlendMode#blend_mode which can be called with or without a block
* Added a page to the manual
* Added specs

A couple of notes and questions:

* Specs need to be run with pdf-inspector master since the tests need access to BM in the ExtGState (https://github.com/prawnpdf/pdf-inspector/commit/6f0992d421eb64810347f30436506c8a80613d8c)
* The manual page for blend_mode seems like it might not be in the right place. It uses a few items, like images, that aren't mentioned in the manual until much later. Where's the best place to put that?
* Should I rebase the code so changes to transparency.rb don't show up?

Thanks!

iamjohnford:
@packetmonkey and @mogest, if one of you get a chance to review the blend_mode addition it would be great. I'm happy to make any updates. Thanks!

packetmonkey:
@iamjohnford Go ahead and rebase to clean up the diff

I missed it when we merged that change into PDF Inspector but can you please open a new PR to that project with an update to the change log so when we release that gem the new functionality is documented? One that release is done we can then update the gem spec in prawn to require it in this PR.

I'm not too concerned about where the page is in the manual, it makes sense to me but we can change it later if something comes up.

Overall I'm pretty stoked about this pr. As far as correctness I'll need to rely on @mogest and @cheba to confirm the correctness of the new functionality, can I please get a review from you two?

iamjohnford:
@packetmonkey, updated the pdf-inspector change log (https://github.com/prawnpdf/pdf-inspector/pull/20) and rebased this pull request.

packetmonkey:
I just released a new version of the PDF inspector gem and Prawn master now depends on it. Can you rebase on master again which should hopefully make this test suite go green!

iamjohnford:
@packetmonkey Updated!

lib/prawn/graphics/blend_mode.rb
```diff
@@ -0,0 +1,78 @@
...
+        if blend_mode_dictionary_registry[key]
+          dictionary =  blend_mode_dictionary_registry[key][:obj]
+          dictionary_name =  blend_mode_dictionary_registry[key][:name]
+        else
+          dictionary = ref!(
+            :Type => :ExtGState,
+            :BM   => blend_mode
+          )
+
+          dictionary_name = "BM#{next_blend_mode_dictionary_id}"
```
cheba: It seems there's no need for incremental ID tracking. There's only so many values it can take. Let's switch from ID to name-based dictionary names?

```ruby
dictionary_name = "BM#{blend_mode}"
```

Then we can get rid of `name` entry in the registry and `ext_blend_mode_dictionary_id` method.

So it would look like this:

```ruby
def blend_mode_dictionary_name(blend_mode)
  dictionary_name = "BM#{blend_mode}"

  dictionary = blend_mode_dictionary_registry[dictionary_name] ||= ref!(
    Type: :ExtGState,
    BM: blend_mode
  )
  dictionary = blend_mode_dictionary_registry[dictionary_name]

  page.ext_gstates.merge!(dictionary_name => dictionary)
  dictionary_name
end
```

lib/prawn/graphics/blend_mode.rb
```diff
@@ -0,0 +1,78 @@
...
+    # Example:
+    #   pdf.fill_color('0000ff')
+    #   pdf.fill_rectangle([x, y+25], 50, 50)
+    #   pdf.blend_mode(:Multiply) do
+    #     pdf.fill_color('ff0000')
+    #     pdf.fill_circle([x, y], 25)
+    #   end
+    #
+    module BlendMode
+      BLEND_MODES = [:Normal, :Multiply, :Screen, :Overlay, :Darken, :Lighten, :ColorDodge, :ColorBurn, :HardLight, :SoftLight, :Difference, :Exclusion]
```
cheba: Spec mentions four more blend modes: Hue, Saturation, Color, and Luminosity.

Anyway, this constant is not used anywhere in the lib code. Maybe, it should move to the manual where it's used?

cheba:
It looks good. Spec is pretty simple on this one.

lib/prawn/graphics/blend_mode.rb
mogest: Or validate the inclusion of `blend_mode` in `BLEND_MODES`?  I feel happy with either solution.

mogest:
Yeah looks great.  @cheba left one comment in the code that could be addressed.  Brilliant work @iamjohnford.

lib/prawn/graphics/blend_mode.rb
cheba: Validation is not a good solution in this case. Spec mentions that `Normal` is a fallback when renderer can not recognize any of the blend modes listed. This ensures backwards compatibility with newer PDF versions which may introduce new blend modes.

mogest: I guess my concern would be that users of prawn are likely to stuff it up, for example use `:multiply` instead of the un-ruby-looking `:Multiply`, or even `'multiply'`, and then wonder why it's not working properly.  That will probably happen a lot more frequently than the PDF spec adding a new blend mode which would then manually need to be added to the list, being an annoyance to someone who wanted to use the new mode soon after its release.

I can see the logic in your proposal though; it's probably more a style/approach question for the project, which I will certainly defer to you :bow:

iamjohnford:
I updated the code with the following:

* Slight change to the way the dictionary name is created when arrays are passed
* Moved BLEND_MODES into the manual
* Updated the code and manual example to include :Hue, :Saturation, :Color, and :Luminosity. Acrobat Reader can display these types but OS X Preview doesn't seem to.

I rebased an it's ready for review.

lib/prawn/graphics/blend_mode.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      private
+
+      def blend_mode_dictionary_registry
+        @blend_mode_dictionary_registry ||= {}
+      end
+
+      def blend_mode_dictionary_name(blend_mode)
+        key = blend_mode.is_a?(Array) ? blend_mode.join : blend_mode
```
cheba: You can make it a bit simpler.

```ruby
key = Array(blend_mode).join('')
```

Basically, you wrap the input into array if it's not already and then join the values. Also you should always specify the joiner string unless you're dealing with the data from `ENV`.

cheba:
@iamjohnford This is great. :star:

iamjohnford:
@cheba, Thanks! I updated the code based on your feedback.

iamjohnford:
@packetmonkey @cheba Anything else you need me to do for this PR?

packetmonkey:
@iamjohnford I'm slowly burning though my prawn backlog, I'll look at this one next.

Thanks again for the help!

packetmonkey:
@iamjohnford You are **AWESOME**

Thanks so much for the PR (and the patience waiting for us to get this merged in!!!!)

Because you got a PR merged, you get commit access to prawn :tada: :fried_shrimp: 

[Here are the details](https://github.com/prawnpdf/prawn/wiki/Contributor-welcome-notes) on what that means.

Seriously, thanks again for the PR. Honest real smiles over here :smile:

Thanks @mogest and @cheba for the extra :eyes: on this. Internet high fives all around :hand:

iamjohnford:
Thanks so much! 😁👍

Issue #783, opened 2014-10-09
Palleted PNG w. transparency unsupported


mfly:
I was creating some PDF files with PNG images and some of them came out with transparency issues. After some tests I narrowed the issue to palleted PNG files. Unfortunately, most of images I use are palleted so it was a real problem.

Quick look at the code reveals that transparency is completely broken for palleted files:
https://github.com/prawnpdf/prawn/blob/master/lib/prawn/images/png.rb#L211

Then I created a test script and a set of test images (https://dl.dropboxusercontent.com/u/20390548/prawn-png.tar.bz2). As expected, everything seems to work, except the palleted PNG with transparency.

Maybe it would be a better idea to use some external library than to try implement PNG handling inside prawn?

Suggested workaround for issues with transparency in PNG files is to use an external tool to convert files into rgba/grayscale mode before using them with prawn.

For me, the best solution is flattening the files with white background and converting it to 8-bit-per-channel RGB with the following command:
```
convert #{file} -background white -flatten -define png:color-type=2 -define png:bit-depth=8 #{out_file}
```

edborden:
@mfly What does the `convert` method refer to in your workaround? It doesn't look like it's part of Prawn, or I can't find mention of it.

practicingruby:
@edborden: The `convert` utility is provided by ImageMagick, and is run on the command line. It's not part of Prawn.

Siyfion:
I know it's been a while since this was last discussed, but what is the currently recommended work-around for this? Does `prawn-fast-png` fix this issue? or `prawn-gmagick`? Is it ever going to be supported natively in Prawn?

Fryie:
While I understand the "no external runtime dependency" stance, it makes little sense IMHO to provide a broken image handling implementation instead. Architecturally it would be more sound to either remove the default PNG image handling or to at least raise a warning when using it.

practicingruby:
@Fryie: I agree, and already mentioned this in the current thread.

> It seems like for now we should think about raising a NotImplementedError for palleted PNGs w. transparency, and then look towards fixing the problem.

practicingruby:
@packetmonkey I marked this as a bug... because as @Fryie said... broken behavior without a warning is pretty much a defect. Consider fixing in a future release. :smile:

Siyfion:
Just to chime in, I can confirm that adding the `prawn-gmagick` plugin fixes this issue for me.

Fryie:
[...]

PR #920, opened 2015-11-21
Added support for transparency in indexed PNG images

cheba:
Solves #783

Here's how `palette-8.png` (from the supplied test case) looks like in the resulting document.

screen shot 2015-11-21 at 11 12 46 pm


packetmonkey:
Nice! In order to get the test suite passing, do we need to just remove the spec that was ensuring we raised an exception for palleted PNG files? Also I think you left a [puts](https://github.com/prawnpdf/prawn/pull/920/files#diff-857d67def6a8ff6448b08d3211b367c0R157) in the specs, we should remove that before merging.

cheba:
@packetmonkey My bad. Fixed specs, removed `puts`, and added a Changelog entry.

danielgrippi:
this fixes an issue i've been seeing on my end.  when can we expect to see this merged?

sublimecoder:
This would fix and issue I have. When can this be merged?

packetmonkey:
@cheba Strangely master is now broken in Travis, I re-ran the build and it looks still broken after merging this and it bundles locally correctly. Can you check this out to see if we broke something with this PR?

PR #922, opened 2015-12-07
Formatted text has wrong leading with one line paragraphs

maerch:
_What's wrong_: If you are indenting a paragraph, but have only one line to print, it will be printed by `draw_indented_formatted_line` completely. Thus, `remaining_text` is empty, but Prawn tries to print it anyway. The problem is that `fill_formatted_text_box` is still changing the `y` value of the document if you specify leading:

``` ruby
      self.y -= box.height
      self.y -= box.line_gap + box.leading if @final_gap
```

`box.height` and `box.line_gap` are zero for an empty remaining text, but box.leading is not. This causes a wrong leading in your text if you have single line paragraphs.

_Example_:

``` ruby
require "prawn"

Prawn::Document.generate("too_much_leading.pdf") do
  text_options = {
    indent_paragraphs: 30,
    leading: 20
  }
  text "Hello World!", text_options
  text "Second paragraph on more than one line!"*5, text_options
  text "Third paragraph on more than one line!"*5, text_options
end
```

Space between the first paragraphs is too large. It should have the same leading as between the second and third paragraph.

### DIFF
--- a/lib/prawn/text.rb
+++ b/lib/prawn/text.rb
@@ -208,8 +208,10 @@ module Prawn
             end
           end
 
-          remaining_text = fill_formatted_text_box(remaining_text, options)
-          draw_remaining_formatted_text_on_new_pages(remaining_text, options)
+          unless @all_text_printed
+            remaining_text = fill_formatted_text_box(remaining_text, options)
+            draw_remaining_formatted_text_on_new_pages(remaining_text, options)
+          end
         end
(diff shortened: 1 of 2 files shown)


pointlessone:
@maerch Thank you for your contribution. I've rebased this PR on master and merged it but GitHub didn't pick it up. I will close this PR now.

PR #926, opened 2015-12-17
Install latest bundler before bundling the build

cheba:
We have a few failing builds on master because of conflict between new rubygems and older bundler.

This PR fixes build of master branch after merge of #920.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ rvm:
   - 2.1
   - 2.2
   - rbx-2
-  - jruby
+  - jruby-1.7.22
 env:
   JRUBY_OPTS=--2.0
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,12 @@ before_install:
   - gem install bundler
   - bundle --version
 rvm:
-  - 2.0
-  - 2.1
-  - 2.2
+  - 2.0.0-p0
+  - 2.0.0-p648
+  - 2.1.0-p0
+  - 2.1.8
+  - 2.2.0-p0
+  - 2.2.4
   - rbx-2
   - jruby-9.0.4.0
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: ruby
+before_install:
+  - gem install bundler
+  - bundle --version
 rvm:
   - 2.0
   - 2.1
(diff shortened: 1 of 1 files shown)


cheba:
Bumped JRuby to 1.7.22 (the one before the latest 1.7). It fixed bundler installation.

We probably should specify versions for all the other implementations, too. Travis updates RVM and rubies once in a while and that changes our build matrix. What was green yesterday may fail today.

PS: JRuby build times are ridiculous. 9m+ vs >1m for other implementations.

cheba:
Also we probably should start testing on JRuby 9k.

packetmonkey:
I agree we should add 9k to travis, can you do that? Then we can get the rest of the ready PRs merged.

I'm not sure I mind automatically using the latest version of a supported branch in ruby. I think the logical alternative would be to explicitly label every patch release of ruby we support, which would give us a massive build matrix. Plus we would have to remember to update travis every time there is a new release, which means eventually it will get forgotten.

Thoughts?

cheba:
The problem with Travis is that they don't use the latest stable release of Ruby. Say, 2.2 means 2.2.0p0 while latest would be 2.2.4 (or 2.2.3 as of the time of this PR submission).

We should test on the oldest supported version of each supported implementation and each latest minor release, and probably ruby-head (or head of each supported implementation).

This indeed would give us much bigger build matrix but also prevent breakage by Travis env updates.

packetmonkey:
Ah! I thought they automatically pointed at the latest version.

I looked at [the travis docs](https://docs.travis-ci.com/user/languages/ruby/#Choosing-Ruby-versions-and-implementations-to-test-against) and they do mention that the aliases change but they don't necessarily guarantee the latest version.

So I think you are totally right, lets track the oldest and newest of the branch to keep our build count lower but hopefully cover any edge cases in ruby.

Can you get that set up?

cheba:
I'll try. Commits here soon.

cheba:
@packetmonkey Updated.

kriztynna:
@cheba Thanks for the transparency fix and for this one too. @packetmonkey When do you expect these changes will be pushed to rubygems.org?

packetmonkey:
@kriztynna I hope to get a release cut within the next week or so, by the end of the year certainly.

kriztynna:
@packetmonkey Thanks for the update! Will keep an eye out for the release.

PR #929, opened 2015-12-19
Fixed Rubygems licenses

cheba:
* Make it clear that Prawn license is a bit different to Ruby license
* Use recommented names for GPL


packetmonkey:
Thanks!

PR #931, opened 2015-12-23
Update to RSpec 3

cheba:
This is mostly a mechanical changes.

* Replaced all `.should` with `expect()`
* Removed deprecated options
* Unrolled some blocks
* Cleaned up everything that RSpec complained about


packetmonkey:
Nice, Thanks!

PR #933, opened 2015-12-26
Handler exception when transparency is not present

hoanghiep90:


cheba:
Hello @hoanghiep90 and thank you for contribution.

Could you please provide a spec for this change?

kriztynna:
Resolved with #938.

Issue #937, opened 2016-01-13
Generating a new pdf document fails when there is an indexed png without transparency


kriztynna:
Generating a pdf using indexed png images with transparency works, but if the images do not have transparency, it fails. Submitting a PR shortly with the fix and related spec.

```
#<NoMethodError: undefined method `[]' for nil:NilClass> with backtrace:
         # ./lib/prawn/images/png.rb:301:in `block in generate_alpha_channel'
         # ./lib/prawn/images/png.rb:295:in `upto'
         # ./lib/prawn/images/png.rb:295:in `generate_alpha_channel'
         # ./lib/prawn/images/png.rb:112:in `split_alpha_channel!'
         # ./lib/prawn/images/png.rb:144:in `build_pdf_object'
         # ./lib/prawn/images.rb:93:in `build_image_object'
         # ./lib/prawn/images.rb:66:in `image'
         # ./spec/acceptance/png_spec.rb:31:in `block (6 levels) in <top (required)>'
         # ./lib/prawn/document.rb:226:in `instance_eval'
         # ./lib/prawn/document.rb:226:in `initialize'
         # ./lib/prawn/document.rb:142:in `new'
         # ./lib/prawn/document.rb:142:in `generate'
         # ./spec/acceptance/png_spec.rb:23:in `block (5 levels) in <top (required)>'
         # ./spec/acceptance/png_spec.rb:22:in `block (4 levels) in <top (required)>'
```

cheba:
It looks like it's the same issue as #933. Is this the case?

kriztynna:
@cheba Yes same issue, different PR.

cheba:
@kriztynna Good. Looking forward to it.

kriztynna:
Resolved by #938

PR #938, opened 2016-01-13
Fixes bug preventing generation of pdfs with non-transparent pngs

kriztynna:
1. Fixes bug
2. Builds out feature spec that would have caught this bug

Solves https://github.com/prawnpdf/prawn/issues/933 and https://github.com/prawnpdf/prawn/issues/937

Already green on Travis: https://travis-ci.org/kriztynna/prawn


cheba:
Thank you for your contribution. It's much appreciated.

However, I think it's not an optimal solution. For instance, In case of indexed PNG without transparency this code would still generate a mask for the image. A completely opaque mask. It's a redundancy that doesn't contribute to appearance of a document, makes a document heavier and rendering more complicated.

I would suggest completely skip mask generation when there's no transparency. I believe this can be achieved by changing `alpha_channel?` method.

kriztynna:
Thanks @cheba for your prompt and thoughtful response. I've amended the PR per your feedback, almost exactly in the way you suggested. Please check it out when you have a moment.

lib/prawn/images/png.rb
```diff
@@ -108,7 +108,7 @@ def colors
       #
       def split_alpha_channel!
         if alpha_channel?
-          if color_type == 3
+          if color_type == 3 && !@transparency.empty?
```
cheba: `@transparency.any?` seem a bit shorter and simpler.

But that's just nitpicking.

cheba:
@kriztynna Thank you. This is pretty much what I had in mind.

kriztynna:
Thanks @cheba. I took a closer look at the output of the amended commit, and actually changed it back to something more like the original pull request. While your concerns about performance and such are justified, I think you'll find the difference in output is well justified. For indexed pngs without a transparency palette, we're still going to need to `generate_alpha_channel`, otherwise the output will be garbled, as the attached example files demonstrate.

[Type 3 with @transparent.any? method.pdf](https://github.com/prawnpdf/prawn/files/90671/Type.3.with.transparent.any.method.pdf)
[Type 3 with current PR.pdf](https://github.com/prawnpdf/prawn/files/90672/Type.3.with.current.PR.pdf)

I will say this about the original PR: it didn't handle type 3 pngs with transparency correctly (made them opaque, as you might have surmised). This version no longer does; I've manually checked the output of the acceptance specs for each kind of file, and it's the best solution I've been able to come up with that handles both Type 3's (with and without transparency) correctly.

kriztynna:
Ok, realize I've gone back and forth on this a couple of times, but it's been a learning process. Think I finally understand what you meant about targeting `alpha_channel?` to fix this bug, and amended the commit accordingly.

The attached pdf shows how the current PR would print indexed pngs. It _now_ matches what prawn would have done before the transparency fix was implemented (the last version of my PR was close, but no cigar). And no unnecessary mask generation to boot.

[Type 3 without generating alpha channel.pdf](https://github.com/prawnpdf/prawn/files/91186/Type.3.without.generating.alpha.channel.pdf)

Hope that does the trick.

packetmonkey:
@cheba I'm not as familiar with the png spec as you (or at all really) so once you are good with the branch go ahead and merge it.

Thanks for the help @kriztynna!

packetmonkey:
@kriztynna Because you had a pull request merged, you get commit access! :tada:

[Here are the details.](https://github.com/prawnpdf/prawn/wiki/Contributor-welcome-notes)

Thanks again for helping prawn!!

kriztynna:
Thanks @packetmonkey, I wasn't expecting that! :dancer: 
And thank you to @cheba for reviewing the PR!

kriztynna:
Resolved #938

kriztynna:
Resolved #937

PR #954, opened 2016-03-10
Ruby versions update

pointlessone:


mojavelinux:
I'm okay with these decisions, but I would like to ask that Prawn be more sensitive about dropping support for Ruby versions that are still in heavy use. I say that because it put us in a really bad spot when support for Ruby 1.9.3 was dropped *before* JRuby 9000 was finished (and stable). This ended up being a huge support cost and a lot of extra work to try to get it running under JRuby 1.7 when JRuby 9000 wasn't a drop-in replacement. These users matter and we should be sensitive to their upgrade path. Just because MRI moves forward doesn't mean all the Ruby implementations do.

pointlessone:
@mojavelinux I understand your argument. Creating inconvenience for users is a bad thing. I totally agree with you.

Yet, I think you're making the issue bigger than it actually is. We officially started supporting JRuby in Prawn 1.0. [[1]](https://groups.google.com/forum/#!searchin/prawn-ruby/jruby/prawn-ruby/8Iv0L_HeHDo/EaxZ1CjkGlYJ) [[2]](https://groups.google.com/forum/#!searchin/prawn-ruby/jruby/prawn-ruby/PwxhCiJgfcI/yWXx0FP7vD4J) At that time it was JRuby 1.7 in 1.9 mode.

In Prawn 2.0 we dropped 1.9 mode and required JRuby 1.7 2.0 mode. [[3]](https://groups.google.com/forum/#!searchin/prawn-ruby/jruby/prawn-ruby/65_lsabpzbw/qp0zb5Vy5ysJ). Not that that thread has not a single comment on the change. I tried to find evidence of that big outcry caused by JRuby 1.7 1.9 mode but I found nothing. Also note that full 2.0 mode never really come true in JRuby 1.7. Which is shame but is unrelated to this discussion.

Now some stats about JRuby usage. There's no definite data as for how many people actually using Prawn with JRuby and JRuby 1.7 specifically. Though, I gathered some numbers that might give an estimate.

Main (this) Prawn repository has 607 closed, and 44 open issues (651 total). Only 11 of them mention JRuby. That's about 1.7%.

Groups has 1753 threads. Only 27 threads mention JRuby. That's about 1.5%.

Here's [last year stats from Semaphore](https://semaphoreci.com/blog/2015/12/01/ruby-version-usage-in-commercial-projects-2015-edition.html). As you can see they show JRuby (all versions) at around 1%. Morover, that number seem to be constant over last three years. And is applicable to a wider variety of projects. Yet, it's mostly consistent with our numbers.

1% is a small number. Yes, they still are Prawn users and are still valued. But our resources are limited. We are willing to work towards better product for them with them. But we also can not spread the little time we have to support old Ruby versions. If those users stayed with older Ruby for whatever reason they likely understood the trade-offs.

mojavelinux:
I want to emphasize that I'm looking towards the future, not dwelling on the past. I'm requesting more careful consideration for Ruby version upgrades from this point forward.

You are right, there wasn't a big outcry when the supported JRuby version was changed. As a result, we ended up in a tight spot in the AsciidoctorJ project where no combination of Ruby/Prawn versions worked. From that experience, we've learned that we need to speak up. And that's exactly what I'm doing here.

> In Prawn 2.0 we dropped 1.9 mode and required JRuby 1.7 2.0 mode.

What we subsequently learned is that 2.0 mode in JRuby 1.7 is severally incomplete. It's unfortunate, but the transition from JRuby 1.7 to JRuby 9.0.0.0 was a unique situation and is now behind us. We probably don't have to worry about that kind of colossal leap again now that JRuby is tracking MRI more closely. Still, we should consider when JRuby drops support for a version of Ruby, not just MRI.

> If those users stayed with older Ruby for whatever reason they likely understood the trade-offs.

The problem is, there was no where for JRuby users to go. JRuby 9.0.0.0 wasn't ready (and is still stabilizing), while 2.0 in JRuby 1.7 wasn't complete. It's very unfair to say that they intentionally stayed on an older Ruby. They were stuck.

> Main (this) Prawn repository has 607 closed, and 44 open issues (651 total). Only 11 of them mention JRuby. That's about 1.7%.
> Groups has 1753 threads. Only 27 threads mention JRuby. That's about 1.5%.

I think you are grossly underestimating the number of Prawn users you have using JRuby. Most of the big users of Asciidoctor are using Prawn via AsciidoctorJ PDF, and adoption is broadening. That includes all of Pivotal, many divisions of Red Hat, MuleSoft, many Eclipse projects and more (see https://github.com/asciidoctor/asciidoctor.org/issues/270).

However, you won't see any of those users mentioning JRuby in the Prawn channels. Instead, those issues come in through Asciidoctor (via AsciidoctorJ). My voice represents all those users. And I hope you consider them to be as valuable as any one using Prawn directly.

mojavelinux:
Regarding the numbers, I want to clarify that there is a difference between a) users that choose to use JRuby over other Ruby implementations and b) users that indirectly use JRuby because they are using Prawn via a Java application (or build tool). My point is that you aren't counting the latter group, which I think is a much bigger group than the former.

robertpanzer:
Hi,

my 2 cents on this:

> Now some stats about JRuby usage. There's no definite data as for how many people actually using Prawn with JRuby and JRuby 1.7 specifically. Though, I gathered some numbers that might give an estimate.

> Main (this) Prawn repository has 607 closed, and 44 open issues (651 total). Only 11 of them mention JRuby. That's about 1.7%.

I seriously question these numbers.
I think the user base of asciidoctorj-pdf is huge!
Every project that renders PDF from Asciidoctor using the asciidoctor-maven org asciidoctor-gradle plugin will use the combination of Prawn and JRuby.
And I think that users who have issues file them against asciidoctor-pdf, so that these issues never reach the prawn project.

Switching to JRuby 9k is a No-No at the moment because JRuby 9K doesn't work reliably on Windows yet and you will always have some guys in a project using Windows. 
(Every version of JRuby 9K has different issues on Windows, but there's no version that runs fine)

On the other hand using the old version 1.3 of Prawn still has an issue that prevents setting a background image for a title page.
This makes asciidoctorj-pdf not being able to provide a professional documentation even though Dan has made such an awesome work on asciidoctor-pdf.

To sum it up I would really like to see a Prawn version that still runs on JRuby 1.7.

(I hope I didn't sound too complaining, but as one of the devs of AsciidoctorJ-PDF I am really passionate about this:) )

Cheers
Robert

practicingruby:
The versions of Prawn that work on JRuby 1.7 will continue to work, and if you're unable to move to JRuby 9000, please use those.

It has always been my intention to support whatever versions of Ruby are supported by the core Ruby team, for as long as they're supported. That has more-or-less been applied consistently throughout the history of the project, and that's what I'd like to see continued.

So we will continue to support whatever version of JRuby that overlaps with the the oldest officially supported versions of Ruby provided by ruby-core. If JRuby trails behind, that means you'll need to trail behind on Prawn versions, but that as soon as JRuby "catches up", you'll be able to catch up on your Prawn versions too.

So, that's basically the end of the story. Please adjust expectations accordingly.

practicingruby:
@pointlessone: Please update this part of the README when you get a chance:

> Because Prawn is pure Ruby and all of its runtime dependencies are maintained by us, it should work pretty much anywhere. We officially support MRI (2.1, 2.2, and 2.3) and JRuby 9000, however we will accept patches to fix problems on other Ruby platforms if they aren't too invasive.

There was a period in Ruby's history where supporting the lowest core Ruby supported version meant also that things would work on stable versions of JRuby and Rubinius. This isn't completely true anymore, so we should make it clear that our support for alternative implementations is limited by their compatibility with officially supported Ruby versions.

mojavelinux:
> The versions of Prawn that work on JRuby 1.7 will continue to work

The point is that they don't because Prawn forces Ruby 2 at installation time and that breaks JRuby environments using JRuby 1.7. It's just not compatible.

You've made your position clear, though. We simply have to say that the only way to use Prawn, and hence Asciidoctor PDF, via JRuby/Java is to be at the bleeding edge where things may not work as expected on all platforms. I can't stop the support requests from coming in (and believe me, they will keep rushing in), but I can at least say it's out of our control. C'est la vie.

practicingruby:
> The point is that they don't because Prawn forces Ruby 2 at installation time and that breaks JRuby environments using JRuby 1.7. It's just not compatible.

No, they don't. Not if you lock to the versions that support Ruby 1.9.3.

> You've made your position clear, though. We simply have to say that the only way to use Prawn, and hence Asciidoctor PDF, via JRuby/Java is to be at the bleeding edge where things may not work as expected on all platforms. I can't stop the support requests from coming in (and believe me, they will keep rushing in), but I can at least say it's out of our control. C'est la vie.

To clear up any misconceptions, we don't give Asciidoctor or its users special treatment, and we also don't run this project as a democracy. You will have the same voice in PrawnPDF with 50 users as you would with 500 or 500,000 or 500,000,000.

I understand that our policies might be incompatible with your goals. If that's the case, please do create your own distribution of PrawnPDF. That would be fine!

headius:
A few clarifications:

> Main (this) Prawn repository has 607 closed, and 44 open issues (651 total). Only 11 of them mention JRuby. That's about 1.7%.
> 
> Groups has 1753 threads. Only 27 threads mention JRuby. That's about 1.5%.

JRuby almost always hides well as a subset of Ruby users, because most JRuby users don't report they're JRuby users until something is obviously a JRuby issue. I would *expect* JRuby's mention count to be low in the issue tracker for a pure-Ruby project.

JRuby users also have some very fast, very complete JVM-based PDF libraries available to them, like itext, and obviously those users won't show up here.

> Here's last year stats from Semaphore

These stats are always a bit questionable coming from one company, since they only reflect apps deployed or connected to that company's systems in some way. Those results will only reflect the composition of a very specific subset of users. In this case, 2014-2015 was obviously a big time for people (using Semaphore) transitioning to 2.x versions of Ruby, which JRuby did not support. That's not the case anymore.

> What we subsequently learned is that 2.0 mode in JRuby 1.7 is severely incomplete.

JRuby 1.7's 2.0 mode was always considered (and declared) experimental and it only implemented a few rudimentary Ruby 2.0 features. It took a while to get JRuby 9k out there, but we're now only a few months behind CRuby. JRuby 9.1 will be out in the next few weeks supporting Ruby 2.3, which was only released a couple months ago.

> Switching to JRuby 9k is a No-No at the moment because JRuby 9K doesn't work reliably on Windows yet and you will always have some guys in a project using Windows.

Let us know what issues are blocking a move to 9k and we'll prioritize them. @enebo has been working on getting Windows compat back where it should be for next releases.

> The "bleeding edge" of JRuby is unfortunately the trailing edge of core Ruby.

This is no longer true at all. The latest released JRuby (9.0.5) supports Ruby 2.2 and we'll have a release of JRuby 9.1 with Ruby 2.3 support shortly.

--

Finally, I just want to reiterate that if any issues keep folks from moving to 9k, we need to know those issues are blockers so we can fix them. We're only planning on supporting the 1.7 line until the end of the year, and 9k+ is the way of the future.

practicingruby:
@headius: Thanks for sharing these details.

It seems like as long as we continue to support JRuby 9k, these problems will go away for users as soon as they move away from JRuby 1.7, hopefully for good.

As long as JRuby 9k overlaps with at least one officially supported Ruby version, we'll continue to build against it and report problems upstream when they are reported to us.

enebo:
@practicingruby we will have a 2.3 compatible release out this month too.  So our overlap window keeps getting smaller wrt C Ruby.  2.0 support itself took longer because we completely replaced our runtime system so it was a big one-off cost which made the 1.7 -> 9k transition much longer than it should have been.  It is unfortunate that cost has caused some friction...

Issue #962, opened 2016-06-04
Document#start_new_page shouldn't mutate its options


johncip:
We started freezing constants recently [to satisfy Rubocop](http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/MutableConstant).

Our usage is similar to:

```ruby 
class ReportRenderer
  MARGINS = { margin: 54.0, bottom_margin: 72.0 }.freeze

  def initialize
    @pdf = Prawn::Document.new(MARGINS)
    # ...
  end

  def add_contents
    @pdf.start_new_page(MARGINS) # RuntimeError: can't modify frozen Hash
    # ...
  end
end
```

(In reality, we set the margins on each new page because sometimes we change them for a particular page based on its content, but a class constant seems like the right place to put the default ones.)

In any case, mutating an option hash seems to violate principle of least surprise (at least, we were surprised when our PDFs broke 😄 ) and the fix is straightforward enough.

The modification happens here, in `apply_margin_options`:
https://github.com/prawnpdf/prawn/blob/master/lib/prawn/document.rb#L693

This can serve as a test case:

```
Prawn::Document.new.start_new_page({margin: 0}.freeze)
```

johncip:
Hey @crazymykl -- made a PR here, if it helps: https://github.com/prawnpdf/prawn/pull/963

johncip:
Hey, when's the appropriate time to close these? After the PR makes it in, or after the next release?

pointlessone:
@johncip Don't worry about that. We do issue triage once in a while and clean up these resolved issues.

PR #963, opened 2016-06-04
Don't mutate options hash in Document#start_new_page

johncip:
Currently, usage like this:

```ruby
Prawn::Document.new.start_new_page({margin: 0}.freeze)
```

will raise:

```ruby
RuntimeError: can't modify frozen Hash
```

The mutation happens in `apply_margin_options`, and only when the `:margin` key is present. This PR refactors that method so that it doesn't need the options hash for temporary storage.

See more details here: https://github.com/prawnpdf/prawn/issues/962

### DIFF
--- a/lib/prawn/document.rb
+++ b/lib/prawn/document.rb
@@ -683,6 +683,8 @@ module Prawn
     end
 
     def apply_margin_options(options)
+      options = options.dup
+      
       if options[:margin]
         # Treat :margin as CSS shorthand with 1-4 values.
         margin = Array(options[:margin])
--- a/lib/prawn/document.rb
+++ b/lib/prawn/document.rb
@@ -692,11 +692,8 @@ module Prawn
                     0 => [] }[margin.length]
 
       sides.zip(positions).each do |side, pos|
-        state.page.margins[side] = margin[pos] if pos
-
-        # side-specific margin options take precedence over :margin
-        specific = options["#{side}_margin"]
-        state.page.margins[side] = specific if specific
+        new_margin = options["#{side}_margin"] || (margin[pos] if pos)
+        state.page.margins[side] = new_margin if new_margin
       end
     end
 
--- a/lib/prawn/document.rb
+++ b/lib/prawn/document.rb
@@ -684,7 +684,7 @@ module Prawn
 
     def apply_margin_options(options)
       options = options.dup
-      
+
       if options[:margin]
         # Treat :margin as CSS shorthand with 1-4 values.
         margin = Array(options[:margin])
--- a/lib/prawn/document.rb
+++ b/lib/prawn/document.rb
@@ -683,23 +683,20 @@ module Prawn
     end
 
     def apply_margin_options(options)
-      options = options.dup
+      sides = [:top, :right, :bottom, :left]
+      margin = Array(options[:margin])
 
-      if options[:margin]
-        # Treat :margin as CSS shorthand with 1-4 values.
-        margin = Array(options[:margin])
-        positions = { 4 => [0, 1, 2, 3], 3 => [0, 1, 2, 1],
-                      2 => [0, 1, 0, 1], 1 => [0, 0, 0, 0] }[margin.length]
+      # Treat :margin as CSS shorthand with 1-4 values.
+      positions = { 4 => [0, 1, 2, 3], 3 => [0, 1, 2, 1],
+                    2 => [0, 1, 0, 1], 1 => [0, 0, 0, 0],
+                    0 => [] }[margin.length]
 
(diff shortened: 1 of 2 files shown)


crazymykl:
Fix rubocop offenses, then this LGTM.

(I think the "spaces around `{` and `}` are good for when they delimit blocks, but not for simple hashes, but the styleguide predates me.)

johncip:
Whoops, didn't see the rubocop.yml. Done.

pointlessone:
To be honest, I'd rather rewrite the whole method.

Currently there are two loops in it and mutation only happen to carry over "unrolled" margins to the second loop. On top of that we now have a `dup`. This is very inefficient.

If we roll both loops into one we would not need an options dup, we would get rid of extra loop, and we would not need to mutate the options.

johncip:
Something like this?

```ruby
def apply_margin_options(options)
  margin = Array(options[:margin])
  sides = [:top_margin, :right_margin, :bottom_margin, :left_margin]

  # Treat :margin as CSS shorthand with 1-4 values.
  positions = { 4 => [0, 1, 2, 3], 3 => [0, 1, 2, 1],
                2 => [0, 1, 0, 1], 1 => [0, 0, 0, 0],
                0 => [] }[margin.length]

  sides.zip(positions).each do |side, pos|
    state.page.margins[side] = margin[pos] if margin[pos]
   
    # :<side>_margin options take precedence over :margin
    state.page.margins[side] = options[side] if options[side]
  end
end
```

Didn't put it through its paces but it should do the same thing.

@crazymykl  solution also loses the dup: https://github.com/prawnpdf/prawn/commit/2e6e4cf0a767ce91d404e5e1397eb5d964dca91e, but doesn't combine the loop. Honestly, the memory reduction strikes me as more important, anyway.

I'd take the `/modify frozen/` out of your spec, though, Mike -- when you describe the specific error that shouldn't be raised, other kinds get ignored. Seems better to just have it fail on anything.

pointlessone:
@johncip Yes, more or less.

lib/prawn/document.rb
```diff
@@ -683,21 +683,20 @@ def generate_margin_box
...
+      # Treat :margin as CSS shorthand with 1-4 values.
+      positions = { 4 => [0, 1, 2, 3], 3 => [0, 1, 2, 1],
+                    2 => [0, 1, 0, 1], 1 => [0, 0, 0, 0],
+                    0 => [] }[margin.length]
+
+      sides.zip(positions).each do |side, pos|
+        state.page.margins[side] = margin[pos] if pos
+
+        # side-specific margin options take precedence over :margin
+        specific = options["#{side}_margin"]
```
crazymykl: You could use `options["#{side}_margin"] || margin[pos]` as the rhs here, and lose line 695. This avoids calling `[]=` twice on the margins.

johncip: Nice catch. We do still need to guard against `pos == nil` before doing a lookup in the margins array, and default to the last page's margin. So this an option:

```ruby
sides.zip(positions).each do |side, pos|
  new_margin = options["#{side}_margin"] || (margin[pos] if pos)
  state.page.margins[side] = new_margin if new_margin
end
```

While I'm not big on `(... if pos)`, it's not unclear. Giving `positions[0]` dummy values would prevent the presence check, but IMO that *would* be less clear. Overall I do think this is better.

pointlessone: I don't think you need to do `if pos`.

`options["#{side}_margin"] || margin[pos]` would result in `nil` when none is present and the next guard will catch it.

johncip: Unless I'm misunderstanding you -- when there are no margins in the options hash, `pos` is nil each time through the loop. Trying to evaluate `margins[nil]` would raise a TypeError.

Two workarounds I noticed were supplying dummy values to the positions array, or looping over `sides.zip(Array(positions))` -- both seemed less explicit.

pointlessone: You are right. Please disregard my comment.

practicingruby:
I reverted this pull request because it wasn't signed off on by a maintainer.

@crazymykl: Please don't merge into master unless asked to by @packetmonkey or @pointlessone.

@packetmonkey, @pointlessone.... please check this and sign off and/or merge.

crazymykl:
I thought having a commit bit made me a maintainer. I guess I was mistaken.

johncip:
Anyone opposed to a new PR for this, since it can't be re-opened? Would also give me the chance to clean up the commit history a little, if that matters.

pointlessone:
@johncip Please don't worry. I'll handle it this weekend.

PR #966, opened 2016-06-08
Adjust to allow not encodable chars;

gabrielms:
Instead of a error, i think that it's better to replace unknown chars with "?" ...

What you guys think?

### DIFF
--- a/lib/prawn/font/afm.rb
+++ b/lib/prawn/font/afm.rb
@@ -99,6 +99,7 @@ module Prawn
       # is replaced with a string in WinAnsi encoding.
       #
       def normalize_encoding(text)
+        text = remove_not_encodable_chars(text)
         text.encode("windows-1252")
       rescue ::Encoding::InvalidByteSequenceError,
              ::Encoding::UndefinedConversionError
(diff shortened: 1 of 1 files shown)


pointlessone:
I strongly oppose this change. With it you may produce unreadable documents and wouldn't know it.

If you have troubles with encodings you should handle it in the application.

I will close this PR but I'm open for discussion. You are welcome to present your case with more arguments or use cases.

gabrielms:
HI @pointlessone thank you.

I think you probably right. The only thing that bugs me is that prawn throws an error... couldn't that be something else? It's not good to have fatal errors to users because of one strange unhandled char.

But i undestand your point and appreciate it :)

May i please ask you a question? Why prawn doesn't come with a TTF font by default?

pointlessone:
Prawn doesn't come with any fonts because PDF format mandates a few default fonts. Here's a list:
* Times−Roman
* Times−Bold
* Times−Italic
* Times−BoldItalic
* Helvetica
* Helvetica−Bold
* Helvetica−Oblique
* Helvetica−BoldOblique
* Courier
* Courier−Bold
* Courier−Oblique
* Courier−BoldOblique
* Symbol
* ZapfDingbats

For this fonts Prawn has font metrics to properly layout text but the fonts themselves must be available on the target system. That is, they either must be a part of the OS, or must be bundled with a reader.

As for your question if the exception could be something else. Prawn is designed to work in a "streaming" manner. That is, every operation produces some output immediately and there's no way to "rewind" and regenerate a part of the document if you encounter any problems. So, there's simply no way to properly handle anything unexpected. In this case, you have no way to undo text rendering done to this point and so you have no way to correct any errors. Because of that there's no way continuing document rendering. That's why throwing an exception is an acceptable solution in the context of Prawn design.

gabrielms:
Thank you for the explanation @pointlessone :) :+1:

PR #971, opened 2016-07-24
implement respond_to_missing? method for Prawn::View module

Rodrigora:
Hey guys.

first thank you for this awesome gem.

I'm building a library on top of prawn that forwards all `Prawn::Document` methods to `@document` taking advantage of `Prawn::View` module.

My code look like this:

``` ruby

class PdfReport
  include Prawn::View
  # ....
end

report = PdfReport.new

# here is the problem
report.respond_to?(:text)  # => returns false
report.text('Hi there')    # => works
```

The `report` should return `true` since it forwards the `text` method to `@document`. More info: [https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding](https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding)

In this PR I implemented the `respond_to_missing` method for `Prawn::View` module.

Please review and LMK if you agree with this change.
And, however this is a simple change, it may break some codes. I think this should be included in a major version.

Thank you for reviewing it.

### DIFF
--- a/lib/prawn/view.rb
+++ b/lib/prawn/view.rb
@@ -69,6 +69,17 @@ module Prawn
       document.send(m, *a, &b)
     end
 
+    # Checks if +document+ respond_to +method_name+.
+    #
+    # This will correctly returns if Prawn::View object respond or not to a
+    # method:
+    #
+    #   view = Prawn::View.new
+    #   view.respond_to?(:cursor)  # => true
+    def respond_to_missing?(method_name, include_private = false)
+      document.respond_to?(method_name) || super
+    end
+
     # Syntactic sugar that uses +instance_eval+ under the hood to provide
     # a block-based DSL.
     #
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## PrawnPDF master branch
 
+### Implemented `respond_to_missing?` method for Prawn::View
+
+Prawn::View objects now properly returns if it responds to a method or not.
+
+(Rodrigo Ra, [#971](https://github.com/prawnpdf/prawn/pull/971);
+
 ## PrawnPDF 2.1.0 -- 2016-02-29
 
 ### Added support for PNG images with indexed transparency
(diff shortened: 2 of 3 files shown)


packetmonkey:
Thanks for the pr!

First I would like to see an additional spec that would return false to ensure we aren't returning true for all methods, This would work

```ruby
 it "returns false for methods the view object does not respond to" do
    expect(view_object).to_not respond_to(:nonexistant_method)
 end
```

I would like to see an entry in the CHANGELOG documenting the change.

What code do you think this would break? In case someone was relying on `Prawn::View#respond_to?` always returned false?

Rodrigora:
Thank you for your review @packetmonkey 

spec/CHANGELOG added!

re possible breaks, yes I guess you're right.

TY

packetmonkey:
This looks good to me.

@pointlessone Any reason to not merge you can think of?

lib/prawn/view.rb
```diff
@@ -69,6 +69,17 @@ def method_missing(m, *a, &b)
       document.send(m, *a, &b)
     end
 
+    # Checks if +document+ respond_to +method_name+.
+    #
+    # This will correctly returns if Prawn::View object respond or not to a
```
pointlessone: I'm not a native English speaker myself but this doesn't sound right to me.

pointlessone:
@Rodrigora Thank you for the PR.

@packetmonkey Apart form the comment wording all is good.

lib/prawn/view.rb
packetmonkey: Good catch, lets change this to something like 

```ruby
    # Checks if +Prawn::View+ respond_to +method_name+.
    #
    # Returns `true` if the Prawn::View object responds to the method, otherwise
    # returns `false`
```

Does that look good to everyone?

pointlessone:
@Rodrigora Thank you for your contribution. This issue has been resolve in master so I will close this PR now. Feel free to reopen if you feel the solution is inadequate.

PR #978, opened 2016-08-24
Set stroke color correctly

systemed:
Currently the `write_stroke_color` method actually sets the fill color, not the stroke color as it should.

### DIFF
--- a/lib/prawn/graphics/color.rb
+++ b/lib/prawn/graphics/color.rb
@@ -219,7 +219,7 @@ module Prawn
       end
 
       def write_stroke_color
-        write_color(current_fill_color, 'SCN')
+        write_color(current_stroke_color, 'SCN')
       end
 
       def write_color(color, operator)


pointlessone:
Hello Richard,

Thank you for your contribution. It's much appreciated.

Would you please provide specs for the change?

Thanks.

systemed:
Sorry, that's not something I have any experience in. If you like I can close this and open an issue for Prawn not setting the stroke colour correctly.

pointlessone:
No, this is fine. We'll take care of it. Eventually…

Once again, thank you for your contribution. It is much appreciated.

gettalong:
@pointlessone I looked at it and it seems that `#write_stroke_color` is never used. In `spec/graphics_spec.rb` are tests for fill and stroke color, so that area has already tests and should be fine.

As a problem in this area would impact *a lot* of people, I think that `#write_stroke_color` is now probably obsolete.

mojavelinux:
I agree, it seems that `write_fill_color` and `write_stroke_color` are both obsolete. Instead, the lower-level method `write_color` is now used, and only by set_color.

pointlessone:
@systemed Thank you for your contribution. As others have pointed out this method is unused and has been removed in master. I will close this PR now.

PR #980, opened 2016-08-29
Refactor for MRI Regexp bug?

eitoball:
(_This PR is for #961_)

This PR should fix difference between MRI and JRuby on page 61 (`text/line_wrapping.rb`) of manual PDF.

It seems Regexp in MRI has a bug when pattern string contains WINDOWS-1251 encoded string. `Prawn::Text::Formatted::LineWrap#update_line_status_based_on_last_output` sets `@line_contains_more_than_one_word` wrongly when `@fragment_output` contains non-breaking spaces.

Following script shows difference in Ruby 2.3.1 and JRuby 9.0.5.0

```
non_breaking_space = "\xA0".force_encoding('WINDOWS-1251')
soft_hyphen = "\xAD".force_encoding('WINDOWS-1251')

puts '----- WINDOWS-1251 -----'
pat = /\s|[#{soft_hyphen}-]/
p pat.encoding
p non_breaking_space =~ pat

puts '----- UTF-8 -----'
pat = /\s|[#{soft_hyphen.encode('UTF-8')}-]/
p pat.encoding
p non_breaking_space.encode('UTF-8') =~ pat
```

In ruby 2.3.1:

```
----- WINDOWS-1251 -----
0
----- UTF-8 -----
nil
```

In JRuby 9.0.5.0:

```
----- WINDOWS-1251 -----
nil
----- UTF-8 -----
nil
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -34,6 +34,7 @@ desc "Generate the 'Prawn by Example' manual"
 task :manual do
   puts "Building manual..."
   require File.expand_path(File.join(File.dirname(__FILE__), %w[manual contents]))
+  prawn_manual_document.render_file("manual.pdf")
   puts "The Prawn manual is available at manual.pdf. Happy Prawning!"
 end
 
--- a/lib/prawn/graphics/patterns.rb
+++ b/lib/prawn/graphics/patterns.rb
@@ -139,13 +139,14 @@ module Prawn
         x1, y1, x2, y2, transformation = gradient_coordinates(gradient)
 
         key = [
-          gradient.type,
+          gradient.type.to_s,
           transformation,
           x2, y2,
-          gradient.r1, gradient.r2,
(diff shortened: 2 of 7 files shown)


pointlessone:
@eitoball This appears to be very similar to #993. Could you please confirm/refute it?

eitoball:
> This appears to be very similar to #993. Could you please confirm/refute it?

Confirmed. Both are on the same bug, but in different solution. 

#993 states that `While MRI is the standard and JRuby is likely the one who's bug it is, we actually want to use JRuby behaviour.`. JRuby behavior is CORRECT. MRI handles non-breaking-space-character wrongly in all encodings except US-ASCII and UTF-8.

Anyway, #993 has been merged, so I am closing this PR. Thank you for your attention.

PR #985, opened 2016-09-19
resolves #984 don't raise CannotFit if first fragment is a zero-width space

### LINKED ISSUE #984: If text begins with a zero-width space fragment, CannotFit exception is thrown
If a string of text begins with a discrete zero-width space fragment, the line wrap logic incorrectly concludes that the text cannot fit. This happens even when the "wrap by char" option is active.

Here's a scenario that reproduces this problem:

```ruby
require 'prawn'
Prawn::Document.generate 'cannot-fit-bug.pdf' do
  font %(#{Prawn::DATADIR}/fonts/DejaVuSans.ttf)
  bounding_box [0, cursor], width: 50 do
    text %(\u200b<color rgb="0000FF">stringofchars</color>), inline_format: true
  end
end
```

If the leading zero-width space character (`\u200b`) is taken away, the text is rendered without exception.

The problematic line is [line_wrap.rb#L199](https://github.com/prawnpdf/prawn/blob/ec9531b6eb0a61cfed0748a395679f053a2d62cd/lib/prawn/text/formatted/line_wrap.rb#L199) called from [line_wrap.rb#L112](https://github.com/prawnpdf/prawn/blob/ec9531b6eb0a61cfed0748a395679f053a2d62cd/lib/prawn/text/formatted/line_wrap.rb#L112).

A zero-width space fragment is often used as a placeholder for adding non-visible objects such as an anchor point.

mojavelinux:


packetmonkey:
This looks good to me (and kudos for diving into the word wrapping code).

Can you mention the bug fix in the change log?

mojavelinux:
Thanks! I've added an entry to the changelog.

PR #1003, opened 2016-12-22
Fix table headers

stormsilver:
Two things in this PR (I know, I know - I am bad and I feel bad)

1. Since this change https://github.com/prawnpdf/prawn/pull/788 my table headers with `valign: :bottom` are too close to the bottom border (the bottoms of my g's and p's and y's and q's are touching that border. Mama always said to mind your p's and q's - I guess this is what she was talking about. She's a lot smarter than I ever gave her credit for when I was 16). It looks like originally `final_gap` was included in the fix but then removed because... something? Adding it back makes things look good again, but I don't know what other problems this could cause.
2. Since this change https://github.com/prawnpdf/prawn/pull/963 my document margins aren't respected anymore. It turns out that the PR (accidentally?) changed from checking symbol options to checking string options. A single colon fixes the problem.


pointlessone:
@stormsilver Thank you for your contribution.

Could you please provide specs for the fixes so that these bugs would never come back?

stormsilver:
@pointlessone I added a spec for the margins fix, but I was really really really unsure how to test the vertical alignment. I poked around the tests from the original PR but they passed with or without my code change, which seemed... wrong.

PR #1010, opened 2017-03-10
Exposed current_column in ColumnBox

elsurudo:
I need this for the following use-case:

`pdf.bounds.move_past_bottom if pdf.bounds.current_column == 0`

### DIFF
--- a/lib/prawn/document/column_box.rb
+++ b/lib/prawn/document/column_box.rb
@@ -50,6 +50,8 @@ module Prawn
     # work.
     #
     class ColumnBox < BoundingBox
+      attr_reader :current_column
+
       def initialize(document, parent, point, options = {}) #:nodoc:
         super
         @columns = options[:columns] || 3


pointlessone:
Why regular bounding box doesn't work for your use case?

elsurudo:
I have two "parts" of the text, and two columns.

If the first part of the text is shorter than the first column, I want the second part of the text to start at the top of the second column.

If the first part of the text is longer than the first column, I want it to wrap to the second column, and I want the second part to start below the end of the first part.

I wasn't sure how to accomplish this otherwise, but I was able to with the change I made.

pointlessone:
`ColumnBox` exposes `left`. It's the left side of the current column. It's going to be 0 for the first column.

Would that work for you?

pointlessone:
I'm closing this as there's another way of achieving your goal. Feel free to reopen if you think this can be improved to be a universally useful feature.

PR #1013, opened 2017-03-18
Move all requires to top level and use require instead of require_relative

rrosenblum:
Using `require_relative` can cause files to be loaded multiple times. An [example project](https://github.com/wconrad/rspec-duplicate-require) was created to show off this issue with rspec.

Here is the issue being exposed using prawn.
```
rrosenblum@C02NV0KSG3QN:~/work/test$ ls -la
total 32
drwxr-xr-x   8 rrosenblum  staff   272 Mar 18 14:12 .
drwxr-xr-x  44 rrosenblum  staff  1496 Mar 17 18:58 ..
drwxr-xr-x   3 rrosenblum  staff   102 Mar 17 18:59 .bundle
-rw-r--r--   1 rrosenblum  staff     6 Mar 17 18:58 .ruby-version
-rw-r--r--   1 rrosenblum  staff    52 Mar 18 14:12 Gemfile
-rw-r--r--   1 rrosenblum  staff   225 Mar 18 14:12 Gemfile.lock
drwxr-xr-x   3 rrosenblum  staff   102 Mar 17 18:59 vendor
lrwxr-xr-x   1 rrosenblum  staff     6 Mar 17 18:59 vendor2 -> vendor
rrosenblum@C02NV0KSG3QN:~/work/test$ cat Gemfile
source 'https://rubygems.org'

gem 'prawn', '2.2.2'
rrosenblum@C02NV0KSG3QN:~/work/test$ bundle config
Settings are listed in order of priority. The top value will be used.
jobs
Set for the current user (/Users/rrosenblum/.bundle/config): "3"

path
Set for your local app (/Users/rrosenblum/work/test/.bundle/config): "vendor2/bundle"
Set for the current user (/Users/rrosenblum/.bundle/config): "vendor/bundle"

disable_shared_gems
Set for your local app (/Users/rrosenblum/work/test/.bundle/config): "true"

rrosenblum@C02NV0KSG3QN:~/work/test$ bundle exec irb
irb(main):001:0> require 'prawn'
/Users/rrosenblum/work/test/vendor2/bundle/ruby/2.3.0/gems/pdf-core-0.7.0/lib/pdf/core/text.rb:12: warning: already initialized constant PDF::Core::Text::VALID_OPTIONS
/Users/rrosenblum/work/test/vendor/bundle/ruby/2.3.0/gems/pdf-core-0.7.0/lib/pdf/core/text.rb:12: warning: previous definition of VALID_OPTIONS was here
/Users/rrosenblum/work/test/vendor2/bundle/ruby/2.3.0/gems/pdf-core-0.7.0/lib/pdf/core/text.rb:13: warning: already initialized constant PDF::Core::Text::MODES
/Users/rrosenblum/work/test/vendor/bundle/ruby/2.3.0/gems/pdf-core-0.7.0/lib/pdf/core/text.rb:13: warning: previous definition of MODES was here
=> true
irb(main):002:0>
```


pointlessone:
Thank you for contribution but I will decline this PR.

There are 2 reasons:
1) Requires are moved around. I'd rather keep them where they are now.
2) This is not a Prawn issue but user's set up. Also the proposed fix introduces other issues.

Let me expand on the second one.

Ruby doesn't load the same file more than once with `require` (and `require_relative` for that matter). It check if the file is already loaded and skips it if it is. The check is done on a fully expanded file name. `require` and `require_relative` only differ in the way they expand the file name passed to them.

`require` does it by looking in every directory listed in `$:` (also known as `$LOAD_PATH`). It basically iterates through load paths, adds the argument passed to it, adds an extension (".rb", or system's binary library extension such as ".dll", ".so", ".dylib", etc.). When it find one it check if it's already loaded (it's in `$LOADED_FEATURES`) and loads if not.

`require_relative` only differs in the way it looks for a file. It doesn't look in `$:`. It looks in the current directory. The check and load part is exactly the same.

So as you can see, if you have two copies of Prawn (or any other gem) they both are going to be loaded.

Now, the new issue I mentioned is that with only `require` you can only load files from the gem that comes first in `$:`. So if you manage to load a file from the second instance, it will require files from the first one. This is potentially problematic when you have two different versions vendored.

Both approaches have drawbacks. But also both are pathological setups. Ruby (unlike, say, Javascript) has global namespace. It's rather problematic if not downright impossible to have two instances of the same gem loaded at the same time. You should never attempt to do that. If you're working on an app, port your code to a version that is supported by all of your deps. If you're writing a gem, put a proper depspec in your gemspec file.

---

Once again, I really appreciate your time and contribution do not be discouraged by this one time. I will close this PR now. Feel free to reopen if you anything to add to discussion.

pointlessone:
You're invited to participate in the further discussion: #1181.

PR #1045, opened 2017-11-08
resolves #1022 provide more information about missing font

### LINKED ISSUE #1022: Font information missing in UnknownFont error message
When a font cannot be found, users often see the following error message:

```
 is not a known font.
```

This happens when a font family + font style font combination is not found in the font registry. Prawn falls back to loading an AFM font. However, in this case, the name argument is always nil and the error message reports a blank font. All the font information is sitting in the options argument.

In afm.rb, I prose changing:

```ruby
def initialize(document, name, options = {}) #:nodoc:

  unless BUILT_INS.include?(name)
    raise Prawn::Errors::UnknownFont, "#{name} is not a known font."
  end
```

to

```ruby
def initialize(document, name, options = {}) #:nodoc:
  name ||= options[:family]
  unless BUILT_INS.include?(name)
    raise Prawn::Errors::UnknownFont, "#{[name, options[:style]].compact * ':'} is not a known font."
  end
```

That way, the error message will be:

```
Noto Serif:bold-italic is not a known font.
```

The message will be consistent whether it's the family or the family + style that can't be matched.

mojavelinux:
- when an afm font is missing, provide more information to the user
- look for font name in family option in case of missing TTF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## PrawnPDF master branch
 
+### Unknown font message
+
+Provide more detail in error message about unknown font.
+
+(Dan Allen, [#1022](https://github.com/prawnpdf/prawn/pull/1022))
+
 ## PrawnPDF 2.2.2
 
 Relax pdf-inspector depspec.
--- a/lib/prawn/font/afm.rb
+++ b/lib/prawn/font/afm.rb
@@ -49,8 +49,10 @@ module Prawn
       end
 
       def initialize(document, name, options = {}) #:nodoc:
+        name ||= options[:family]
         unless BUILT_INS.include?(name)
-          raise Prawn::Errors::UnknownFont, "#{name} is not a known font."
+          raise Prawn::Errors::UnknownFont,
+            "#{name} (#{options[:style] || 'normal'}) is not a known font."
         end
 
         super
--- a/spec/prawn/font_spec.rb
+++ b/spec/prawn/font_spec.rb
@@ -69,6 +69,14 @@ describe Prawn::Font do
       end
     end
 
+    it 'reports missing font with style' do
+      expect do
+        pdf.font('Nada', style: :bold) do
+          pdf.width_of('hello')
+        end
+      end.to raise_error(Prawn::Errors::UnknownFont, /Nada \(bold\)/)
+    end
+
     it 'calculates styled widths correctly using TTFs' do
       pdf.font_families.update(
         'DejaVu Sans' => {


pointlessone:
@mojavelinux Could you please fix the build?

mojavelinux:
I forgot to run rubocop. I've addressed the violations.

For the record, I have no idea how Travis is running rubocop. When I run it locally, I get 13,000 violations. It would be nice to know what the right command is so I can actually test it locally.

pointlessone:
Default Rake task is running both Rubocop checks and specs. Use bundler to be sure you're running all the right versions.

```bash
bundle exec rake
```

mojavelinux:
Aha, I had the gem extracted under pkg, which was getting caught up in the glob. I think the glob needs to be more specific because it's including stuff it shouldn't.

I took that away and I still get violations. Here's a partial output using Ruby 2.4.1.

```
lib/prawn/graphics/patterns.rb:1:1: C: Missing frozen string literal comment.
require 'digest/sha1'
^
lib/prawn/image_handler.rb:1:1: C: Missing frozen string literal comment.
# ImageHandler provides a way to register image processors with Prawn
^
lib/prawn/security/arcfour.rb:1:1: C: Missing frozen string literal comment.
# Implementation of the "ARCFOUR" algorithm ("alleged RC4 (tm)"). Implemented
^
lib/prawn/encoding.rb:1:1: C: Missing frozen string literal comment.
# Copyright September 2008, Gregory Brown, James Healy  All Rights Reserved.
^
lib/prawn/font_metric_cache.rb:1:1: C: Missing frozen string literal comment.
# font_metric_cache.rb : The Prawn font class
```

mojavelinux:
Strange. I cloned the project again in a fresh space and then the rubocop passed. I have no idea why I'm getting different results.

mojavelinux:
I've tried everything I can think of. I don't know why this old checkout is giving me errors. There are no additional files reported by git. Whatever, I'll just clone it again.

pointlessone:
You should never use a packaged gem for development. Packaged gem used to be a proper distribution mechanism at the very beginning of Ruby gems but since then it has changed (about 14 years ago when RubyForge came about). Nowadays it's actively discouraged to package supporting infrastructure. Packaged gem should only contain files that are required for proper execution and documentation generation.

As for your checkout you may have different version of Rubocop installed. Our config inherits from the default one so somewhat version-dependent.

mojavelinux:
If I run the tests from /tmp/prawn, I get no violations. If I move the prawn directory into my home directory, then I get violations. So it's clearly looking at parent directories, so it's not really a portable set up.

pointlessone:
That's interesting. Rubocop does indeed traverse tree to find config but it should not go all the way tho the `/`. I should check the Rubocop code…

pointlessone:
@mojavelinux Do you have a merge bit on this repo?

mojavelinux:
The problem seems to be that my folder structure is "projects/prawn/prawn". If I rename it to "projects/prawn/prawn-core" then I get no violations. I don't know what the heck is going on there.

mojavelinux:
But on a different computer, I don't get that problem. It seems like information has been cached somewhere. I just don't know where that is. I'll keep poking at it to see if I can figure out what is causing the delta.

mojavelinux:
> Do you have a merge bit on this repo?

It appears so, but merging is blocked unless there is one approved review.

mojavelinux:
> You should never use a packaged gem for development. 

I 100% agree with you. What I meant to say was that I had some junk inside the project left behind when I was looking for something. I should have just said, "I left behind a temporary directory".

But that's a separate issue from why I get violations when the folder structure is "projects/prawn/prawn". But I'm willing to say that it's a dirty cache somewhere.

pointlessone:
> merging is blocked unless there is one approved review

Indeed. My bad. 😅

mojavelinux:
:+1:

mojavelinux:
Regarding rubocop, each computer I try I get different results. If I run `bundle exec rake` I get different results than `bundle exec rubocop`.

mojavelinux:
I added an entry to the changelog. Could you approve again?

mojavelinux:
@pointlessone Did you want me to merge?

pointlessone:
@mojavelinux Yes, please.

mojavelinux:
Done!

PR #1057, opened 2018-03-15
raise an exception when align value is not a Symbol

lucianosousa:


pointlessone:
@lucianosousa Hi! Thank you for your interest in giving back to community!

Could you please give a bit of context for this PR?

lucianosousa:
@pointlessone sure man.
I was trying to align here with `align: 'center'` and the pdf render just broke.
Some rails methods accept both kinds of argument types, eg: https://github.com/rails/rails/blob/master/activemodel/lib/active_model/validations.rb#L256
we can do the same here, don't you think?

pointlessone:
Prawn is not a part of the Ruby on Rails framework. We do not quite share an outlook on API design with Rails.

I agree that it's suboptimal to break document rendering. But I think a proper solution here to let the user know what's wrong instead of extending the API. We generally do not accept strings where only a limited number of named options can be accepted and I don't want to diverge from that in this case.

What do you think of an alternative solution? Add an `else` to that `case` that will raise an `ArgumentError` with a descriptive message.

lucianosousa:
@pointlessone cool. better raise an exception then ;)

pointlessone:
@lucianosousa Would you like to change the code in this PR or l it to someone else?

lucianosousa:
@pointlessone sorry the delay. it's here! ;)

pointlessone:
@lucianosousa I would very much prefer if the exception was risen in every unrecognized case not only when it's not a symbol.

lucianosousa:
sorry being late. updated

pointlessone:
NP. Thank you for keeping working on this.

Could you please put the `raise` into `else` branch of the `case` statement to save us extra check?

lucianosousa:
done @pointlessone sorry being late

lucianosousa:
@pointlessone I guess we need merge #1060 first here

pointlessone:
Not sure. That one improves document consistency but I never saw a difference between 2.2 and 2.3+.

lucianosousa:
@pointlessone just merged and passes! :D ready for u

pointlessone:
@lucianosousa Could you please clean up the history a bit?

lucianosousa:
@pointlessone done

pointlessone:

PR #1065, opened 2018-05-24
Fixing Prawn::View#respond_to_missing? method signature

### LINKED ISSUE #1064: Using Prawn::View mixin causing ArgumentError when calling respond_to?
When you use `Prawn::View` mixin to create custom class that extends Prawn's functionality, the method `respond_to?` gives an error when called with a missing method. E.g.:

```ruby
class Greeter
  include Prawn::View

  def initialize(name)
    @name = name
  end

  def say_hello
    text "Hello, #{@name}!"
  end
end
```
```ruby
greeter = Greeter.new("Gregory")
greeter.respond_to?(:say_hello) # => true
greeter.respond_to?(:non_existing_method)
ArgumentError: wrong number of arguments (given 2, expected 0)
	from (irb):4:in `respond_to?'
	from (irb):4
```

This error is also being raised when calling with some methods that should exist:
```ruby
greeter.respond_to?(:text)
ArgumentError: wrong number of arguments (given 2, expected 0)
	from (irb):4:in `respond_to?'
	from (irb):4
```

This is causing `RSpec` to fail when you try to stub a method for the given object, like:
```
before { allow(greeter).to receive(:text) }
```

vitork:
Fixes #1064

### DIFF
--- a/lib/prawn/view.rb
+++ b/lib/prawn/view.rb
@@ -69,7 +69,7 @@ module Prawn
       document.send(m, *a, &b)
     end
 
-    def respond_to_missing?
+    def respond_to_missing?(m, _include_all = false)
       document.respond_to?(m)
     end
 
--- a/spec/prawn/view_spec.rb
+++ b/spec/prawn/view_spec.rb
@@ -44,4 +44,20 @@ describe Prawn::View do
     view_object.save_as('foo.pdf')
     expect(doc).to have_received(:render_file)
   end
+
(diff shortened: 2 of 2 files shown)


pointlessone:
Could you please squash the commits?

pointlessone:
Thank you very much for your contribution!

vitork:
No problems! Happy to contribute 😄

PR #1066, opened 2018-06-14
Check if image responds to :read

duhast:
Hi! This update allows to source image from "alternate" file implementations like `CarrierWave::Storage::Fog::File` and `CarrierWave::SanitizedFile`.

Very convenient when you're working with images stored remotely.

### DIFF
--- a/lib/prawn/images.rb
+++ b/lib/prawn/images.rb
@@ -137,11 +137,11 @@ module Prawn
 
     def verify_and_open_image(io_or_path)
       # File or IO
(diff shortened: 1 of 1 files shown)


petergoldstein:
@duhast I know it's been a very long time, but do you still have interest in this PR?  If so, any chance you can rebase and add a spec for the "alternate" file implementation case?

duhast:
Oh yeah, it's been a long time 😄 

I have this patch working in production for couple years already, however I don't think I'll have time to work on the spec for PR in near future.

Issue #1073, opened 2018-08-27
character_spacing bug after raising Prawn::Errors::CannotFit


ndbroadbent:
I have some code that catches `Prawn::Errors::CannotFit` and just draws the text normally as a fallback. But there's a bug with `character_spacing`, where it is ignored after the `CannotFit` error,. However, it works if I change `character_spacing` to a new value. This also only happens when the error is raised inside a `pdf.transparent` block.

Here's a minimal POC to reproduce the issue:

```ruby
#!/usr/bin/env ruby
require 'prawn'

Prawn::Document.generate('/tmp/prawn-test.pdf') do |pdf|
  pdf.fill_color('000000')
  pdf.font 'Helvetica', size: 12, style: :normal

  pdf.text_box 'Lorem Ipsum Etcetera',
    at: [55, 633],
    width: 300,
    height: 15,
    overflow: :shrink_to_fit,
    character_spacing: 7

  pdf.font 'Courier', size: 18, style: :normal

  pdf.transparent(0.4) do
    begin
      pdf.text_box 'abc',
        at: [55, 570],
        width: 5,
        height: 5,
        overflow: :shrink_to_fit,
        character_spacing: 7
    rescue Prawn::Errors::CannotFit
      pdf.draw_text 'abc', at: [55, 570]
    end
  end

  pdf.font 'Helvetica', size: 14, style: :normal
  pdf.text_box 'Some More Lorem Ipsum Text',
    at: [55, 550],
    width: 350,
    height: 20,
    overflow: :shrink_to_fit,
    character_spacing: 7
end

`open /tmp/prawn-test.pdf`
```

Result:

<img width="414" alt="screen shot 2018-08-27 at 10 13 10 pm" src="https://user-images.githubusercontent.com/139536/44667970-6ee90e00-aa46-11e8-80bd-c8ae9e983016.png">

You can see that `character_spacing: 7` is not working for "Some More Lorem Ipsum Text".

------------------

If I remove the `pdf.transparent(0.4) do` wrapper by commenting out lines 17 and 28 (but still run the code inside the block), then I see the correct behavior:

<img width="564" alt="screen shot 2018-08-27 at 10 15 30 pm" src="https://user-images.githubusercontent.com/139536/44668111-b96a8a80-aa46-11e8-9f28-7e106444bf2f.png">

[...]

PR #1074, opened 2018-08-27
Clean up document state before raising Prawn::Errors::CannotFit

ndbroadbent:
Fixes a strange `character_spacing` issue that I was experiencing in #1073. All the specs are passing, and this change fixes my POC, but I might have missed something.

### DIFF
--- a/lib/prawn/text/formatted/box.rb
+++ b/lib/prawn/text/formatted/box.rb
@@ -233,6 +233,9 @@ module Prawn
             end
           end
 
+          # Only raise CannotFit error after cleaning up document state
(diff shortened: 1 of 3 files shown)


petergoldstein:
This looks like a mix of two different changes - the first is cleaning up the document state while the second is an option to ignore height overflows.  I'd like to separate these out - it seems likely the first commit will be pulled in more easily than the second.  Will put up a PR with just the first.

petergoldstein:
@pointlessone  This PR should be closed in favor of the underlying fix which was merged into pdf-core here - https://github.com/prawnpdf/pdf-core/pull/53

PR #1079, opened 2018-09-24
Fix indent_paragrahs/no final_gap line spacing bug

jstasiak:
In case of no indent_paragraphs when formatted_text is called it
passes the text to render to Text.fill_formatted_text_box. Then the text
is rendered using the Box class and the Box class in combination with
the Wrap module takes care of spacing between lines. Spacing is present
regardless of final_gap value.

In case of indent_paragraphs set the situation is different.
formatted_text itself prints the first line separately and then the rest
of the text. As such, it has the responsibility to add spacing between
the first line and the rest on its own. When final_gap is set, as it is
by default, it works just fine. When final_gap is not set though, there
was no space between the first line and the second line of the text.

Since formatted_text calls draw_indented_formatted_line to render the
first line and draw_indented_formatted_line calls
fill_formatted_text_box with the single_line flag we can use the flag as
a secondary way to trigger adding a gap.

### DIFF
--- a/lib/prawn/text.rb
+++ b/lib/prawn/text.rb
@@ -382,7 +382,13 @@ module Prawn
       @all_text_printed = box.everything_printed?
 
       self.y -= box.height
-      self.y -= box.line_gap + box.leading if @final_gap
+
+      # If there's no remaining_text we don't really want to treat this line
+      # in a special way, we printed everything we wanted so the special
+      # single_line logic should not be triggered here.
+      if @final_gap || (options[:single_line] && !@all_text_printed)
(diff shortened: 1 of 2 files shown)


jstasiak:
My first approach was buggy in case of single-line paragraphs, I just updated it.

petergoldstein:
@pointlessone Could you please rebase this one?  I think we likely want to merge this once it's green.

pointlessone (approved):

pointlessone:
@jstasiak Thank you for your contribution.

PR #1082, opened 2018-10-11
Address a performance issue and adjust some rubocop names

chills42:


.rubocop.yml
```diff
@@ -12,9 +12,6 @@ RSpec/MultipleExpectations:
 RSpec/NestedGroups:
   Max: 5
 
-RSpec/NotToNot:
```
pointlessone: Why was this removed?

lib/prawn/soft_mask.rb
```diff
@@ -75,7 +75,7 @@ def soft_mask(&block)
         renderer.add_content "/#{soft_mask_registry[registry_key]} gs"
       else
         masks = page.resources[:ExtGState] ||= {}
-        id = masks.empty? ? 'GS1' : masks.keys.sort.last.succ
+        id = masks.empty? ? 'GS1' : masks.keys.max.succ
```
pointlessone: This should be in a separate commit.

.rubocop.yml
chills42: That was accidental, I was actually just fixing that.

PR #1083, opened 2018-10-11
Address a minor performance issue.

chills42:
This is unlikely to have a significant affect, but it should be slightly
more performant while improving code clarity.


pointlessone:
I meant a separate commit, not a whole new PR but it works too, I guess. :)

chills42:
Yeah, I realized that, but also thought it would be a little cleaner this way

pointlessone:

PR #1090, opened 2018-11-23
Correctly handle pathnames by closing opened files again

### LINKED ISSUE #975: Prawn::Images#image method does not close IO object
If a file path is passed to the Prawn::Images#image method (and subsequently the build_image_object method), the IO object opened by verify_and_open_image to read the file is never closed. This prevents Ruby from being able to delete the file path, in the case the file path is a temporary resource.

I understand the motivation for not closing an IO object when it is passed to the Prawn::Images#image method. In that scenario, the IO object is being managed external to the API. However, the Prawn::Images#image should manage (i.e., close) any IO object it itself creates since there's literally no way to access the object to close it.

haslo:
This PR fixes #975 by using a standard `File.open` block for handling path names as images.

The io is `yield`ed within that block instead of returned from the method, and thus is correctly closed after it has been used to read from.


pointlessone:
Wouldn't it be easier to change `verify_and_open_image` to `verify_and_read_image` that would just return read data, instead of io? I imagine, the overall code would be simpler and easier to read as well.

Also, please add specs.

haslo:
@pointlessone Do you know of a way to test open file handles from inside Ruby, without resorting to OS tools? I've added a file double way of testing whether the block syntax is used for File.open, and a version that uses lsof and only runs on systems that use that, but I'm not perfectly happy with either...

lib/prawn/images.rb
```diff
@@ -145,13 +144,14 @@ def verify_and_open_image(io_or_path)
...
         io.binmode if io.respond_to?(:binmode)
-        return io
+        return io.read
       end
       # String or Pathname
       io_or_path = Pathname.new(io_or_path)
       raise ArgumentError, "#{io_or_path} not found" unless io_or_path.file?
-      io = io_or_path.open('rb')
-      io
+      File.open(io_or_path, 'rb') do |file_io|
```
pointlessone:
```ruby
File.read(io_or_path)
```
Instead of the block `open` would work as well.

haslo: It would have to be `#binread` though, right?

pointlessone:
Well, this issue was never considered a serious one because Ruby is quite good a closing files for you. And does so fairly quickly.

It does so when it GCs the file object. I imagine, you can write a reliable spec by preventing that from Happening. Let say, you can disable GC for one spec so that opened file object could live long enough for you to do all the assertion you want.

lib/prawn/images.rb
pointlessone: Yes, you're right. That is closer to the original code.

haslo:
Unfortunately, it happens to be serious for us because the files we open live on an SMB share, who complains about open files fairly quickly. The processes using Prawn are Rails workers in Passenger, who doesn't appear to close the files until the process dies and gets replaced by a new one, which can be minutes...

Both specs are reliable, but I'd prefer to just ask File whether it knows of any open handles. It doesn't look like Ruby has anything of that sort though.

pointlessone:
No, Ruby doesn't keep a list of all open files.

You may be observing this more often in Passenger because many web-servers tune GC to not happen during request. However, it should happen in between requests so I don't know why it has to die to free up file handles.

haslo:
Leaking is of course possible, but the issue arises in the places where we feed Prawn just pathnames as strings. I'm not sure how long it's being retained, the fact is that it's sometimes minutes. We tried to fix the symptoms by always passing file instances instead of filenames, but figured that's silly when we can just help fixing the cause instead of digging through all our code and maintaining lists of files in dozens of places...

pointlessone:
Isn't `ObjectSpace` MRI-specific? I'd like to keep whatever JRuby compatibility we have.

haslo:
I wasn't aware. What do you think of running that particular spec only for MRI?

pointlessone:
I'm not opposed to the idea of impl-specific specs. As long as there are corresponding specs for other implementations.

haslo:
I've rebased it all into one commit. Thanks again for your feedback!

spec/prawn/images_spec.rb
```diff
@@ -43,11 +43,46 @@
...
+
   it 'accepts Pathname objects' do
     info = pdf.image(Pathname.new(filename))
     expect(info.height).to eq(453)
   end
 
+  it 'closes opened files again after getting Pathname objects' do
+    if RUBY_PLATFORM == 'java'
+      # JRuby doesn't support ObjectSpace
+      expect(RUBY_PLATFORM).to eq('java')
```
pointlessone: There needs to be a spec here. Otherwise we can't be sure it works on Ruby.

haslo: I'll try to come up with something... The spec that used `lsof` isn't really great either, that doesn't test anything for Windows platforms. Or would that work for you?

pointlessone: Right now we don't run tests on Windows but we don't have any platform-specific code either. Everything is Ruby-only so in theory we shouldn't have any major issues with adding Windows. It at all possible I'd prefer it stayed that way.

spec/prawn/images_spec.rb
```diff
@@ -43,11 +43,36 @@
...
+
+    expect(file).to_not be_closed
+  end
+
   it 'accepts Pathname objects' do
     info = pdf.image(Pathname.new(filename))
     expect(info.height).to eq(453)
   end
 
+  context 'with File message spy' do
```
pointlessone: This spec feels a bit too coupled to the implementation and doesn't describe the actual intent particularly well but since I can't give any useful advice on how to improve it, let's keep it. However, tag it with issue. Here's an [example](https://github.com/prawnpdf/prawn/blob/2cdd2e7994c742c97f104e2927d2e2fd3dedc767/spec/prawn/view_spec.rb#L48).

haslo: I agree, I don't like testing the implementation either. Something like the version with `ObjectSpace` but working on every platform would've been awesome...

haslo:
I think it would make sense to add the `lsof` and the `ObjectSpace` versions of the specs after all, even if they have no direct equivalents without those tools. The feature the specs test is whether the file handles are closed again. There will then be three specs for that, testing every angle on systems that have all three approaches available, and only testing the implementation-based angle currently in the PR if they don't.

For reference, `lsof` version:

```ruby
  it 'closes opened files again after getting Pathname objects' do
    system_has_lsof = system('lsof -v > /dev/null 2>&1')
    system_has_grep = system('grep --version > /dev/null 2>&1')
    if system_has_lsof && system_has_grep
      gc_was_disabled = GC.disable # make sure GC doesn't remove File instance and close file
      open_files_before = `lsof -c ruby | grep "#{filename}"`
      _info = pdf.image(Pathname.new(filename))
      open_files_after = `lsof -c ruby | grep "#{filename}"`
      GC.enable unless gc_was_disabled
      expect(open_files_after).to eq(open_files_before)
    else
      expect(system_has_lsof && system_has_grep).to be_falsy # just for Rubocop
    end
  end
```

...and `ObjectSpace` version:

```ruby
  it 'closes opened files again after getting Pathname objects' do
    if <guard ObjectSpace availability>
      gc_was_disabled = GC.disable # make sure GC doesn't remove File instance and close file
      open_files_before = ObjectSpace.each_object(File).count { |f| !f.closed? }
      _info = pdf.image(Pathname.new(filename))
      open_files_after = ObjectSpace.each_object(File).count { |f| !f.closed? }
      GC.enable unless gc_was_disabled
      expect(open_files_after).to eq(open_files_before)
    else
      expect(<tautology>).to be_truthy # just for Rubocop
    end
  end
```

If I add these to the PR, even without corresponding specs when the tools aren't available, the logic is tested from three non-perfect angles instead of just one:

* Cross platform compatible, but implementation-based instead of logic-based
* Logic-based, but only works on linux and macOS platforms
* Logic-based, but only works on MRI

If anybody has an idea for a logic-based cross-platform spec after all, I'm all ears. But without that, personally I'd feel happier with these three specs and a comment explaining their raison d'être than just one...

spec/prawn/images_spec.rb
```diff
@@ -43,11 +43,68 @@
...
+      if system_has_lsof && system_has_grep
+        gc_was_disabled = GC.disable # GC of File would close the file
+        open_before = `lsof -c ruby | grep "#{filename}"`
+        _info = pdf.image(Pathname.new(filename))
+        open_after = `lsof -c ruby | grep "#{filename}"`
+        GC.enable unless gc_was_disabled
+        expect(open_after).to eq(open_before)
+      else
+        pending 'Platform dependent spec, only works on linux and macOS'
+        raise 'pending'
```
pointlessone: Is this required? Wouldn't only previous line be enough?

haslo: Rspec doesn't seem to like pending specs that don't fail. IIRC it was introduced in Rspec 3:

screen shot 2018-11-26 at 11 53 19

Error: `Expected pending 'Platform dependent spec, only works on linux and macOS' to fail. No error was raised.`

spec/prawn/images_spec.rb
```diff
@@ -43,11 +43,68 @@
...
+
+      it 'uses only binread, which closes opened files' do
+        # this implicitly tests that a file handle is closed again
+        # because only stubbed binread can be called on not_filename
+        _info = pdf.image(not_filename)
+        expect(pathname_double).to have_received(:binread)
+      end
+    end
+
+    it 'closes opened files, spec with lsof' do
```
fidothe: An easier approach to these kind of specs is to wrap the definition with the platform check, which means you avoid the sometimes-this-spec-is-pending-but-not-always problem:

```ruby
describe "situation" do
  system_has_lsof = system('lsof -v > /dev/null 2>&1')
  system_has_grep = system('grep --version > /dev/null 2>&1')
  if system_has_lsof && system_has_grep
    it 'closes opened files, spec with lsof' do
      ...
    end
  end

  if RUBY_PLATFORM != 'java'
    it 'closes opened files, spec with ObjectSpace' do
      ...
    end
  end
end
```

haslo:
The PR is still in the status "changes requested" ... the specs are as fitting as I think is possible now, right? Is there something more I can / should do?

Issue #958, opened 2016-04-27
Ability to destroy the last or a specific page


sobrinho:
I'm not sure if it's a good idea but would be nice to be able to destroy a page for situations like this:

``` ruby
things.each do |thing|
  # do some rendering

  document.start_new_page
end
```

If you do not want an empty page on the document, you need to perform some check, like this:

``` ruby
things.each do |thing|
  # do some rendering

  if thing != things.last
    document.start_new_page
  end
end
```

If you can destroy a page, you could avoid the check in every iteration, like this:

``` ruby
things.each do |thing|
  # do some rendering

  document.start_new_page
end

document.destroy_last_page
```

Do you think that's too strange or does it makes sense?

andre1810:
For me it makes sense and I would recommend it.

PR #1092, opened 2018-11-30
Add method to delete pages of the document by index

davidsilveira:
https://github.com/prawnpdf/prawn/issues/958

### DIFF
--- a/lib/prawn/document.rb
+++ b/lib/prawn/document.rb
@@ -306,6 +306,26 @@ module Prawn
       end
     end
 
+    # Remove page of the document by index
+    #
+    #   pdf = Prawn::Document.new
+    #   pdf.page_count #=> 1
+    #   3.times { pdf.start_new_page }
+    #   pdf.page_count #=> 4
+    #   pdf.delete_page(-1)
+    #   pdf.page_count #=> 3
+    #
+    def delete_page(index)
+      return false if index.abs > (state.pages.count - 1)
+
+      state.pages.delete_at(index)
+
+      state.store.pages.data[:Kids].delete_at(index)
+      state.store.pages.data[:Count] -= 1
+      @page_number -= 1
+      true
+    end
+
     # Returns the number of pages in the document
     #
     #   pdf = Prawn::Document.new
--- a/spec/prawn/document_spec.rb
+++ b/spec/prawn/document_spec.rb
@@ -163,6 +163,48 @@ describe Prawn::Document do
     end
   end
 
+  describe '#delete_page(index)' do
+    before do
+      pdf.text 'Page one'
+      pdf.start_new_page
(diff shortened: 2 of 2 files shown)


lib/prawn/document.rb
```diff
@@ -306,6 +306,26 @@ def start_new_page(options = {})
...
+    # Remove the last page of the document
+    #
+    #   pdf = Prawn::Document.new
+    #   pdf.page_count #=> 1
+    #   3.times { pdf.start_new_page }
+    #   pdf.page_count #=> 4
+    #   pdf.delete_last_page
+    #   pdf.page_count #=> 3
+    #
+
```
sobrinho: empty line

lib/prawn/document.rb
```diff
@@ -306,6 +306,26 @@ def start_new_page(options = {})
...
+    def delete_last_page
+      last_page = state.page
+      if last_page
+        state.pages.delete_at(state.pages.index(last_page))
+        state.store.pages.data[:Kids].pop
+        state.store.pages.data[:Count] -= 1
+        @page_number -= 1
+      end
```
sobrinho: Use earlier return:

```ruby
def delete_last_page
  last_page = state.page

  return if last_page.nil?

  ...
end
```

sobrinho:
As we discussed, try to abstract to `delete_page(index)` to be able to delete any page instead of only the latest.

pointlessone:
I wonder what is the use case for this feature.

sobrinho:
See #958.

pointlessone:
@sobrinho Thank you.

I'm not convinced that use case warrants an addition to the library.

sobrinho:
@pointlessone this is a pattern that repeats over and over again in all projects that I work.

It's okay if you think that doesn't belong to the core but would be a good helper for us!

pointlessone:
If you want each thing on its own page why don't you start a new page before rendering a thing?

sobrinho:
Because you can't start a new page in the first iteration, otherwise the blank page would be first one instead of the latest.

In that case, the code is like that:

```ruby
things.each do |thing|
  if things.first != thiing
    document.start_new_page
  end

  render
end
```

pointlessone:
I see but I'm still hesitant since there's a straight-forward workaround and the code is not complex or obscure.

sobrinho:
Your call!

Maybe there is other use cases, ours is just this one.

johnnyshields:
Deleting a page at an index is valid request. Ideally it should support negative indexes (`-1` == last page).

The use case given is rather weak/inelegant however.

sobrinho:
Awesome!

pointlessone:
@davidsilveira Please make Rubocop happy. Otherwise it's good.

PR #1097, opened 2019-01-01
Fix transformation stack not being popped

mogest:
save_graphics_state can take a block form which automatically restores
the graphics state when the block completes.  However, the
transformation stack was not being popped in this case.

This fix pops the transformation stack after save_graphics_state is
used in block form.


petergoldstein:
@pointlessone Can you please rebase this?  To me this PR looks like it should be merged.

pointlessone:
@mogest Thank you for your contribution.

Issue #1052, opened 2017-12-05
Release a new version?


Zyphrax:
Could you please release a new version?
The current version was released on March 11th 2017.

The `Remove superfluous pdf-core requires` commit is really useful, silencing a few frustrating warnings. My Gemfile now points to that specific commit, but it is really slow when doing bundle updates.

Zyphrax:
Bump @pointlessone

pointlessone:
I hear you. As soon as I get some time.

shevegen:
We all use prawn when we generate .pdf files in ruby! \o/

lichtamberg:
I hope you have time soon to release a new gem version! 👍

Zyphrax:
Pretty please? 🤗

Zyphrax:
@pointlessone bump

pointlessone:
@Zyphrax Sure. Eventually. Some exciting work’s being landed on supporting gems. Once that’s done expect a new release.

Zyphrax:
@pointlessone New PDF core commit and a few other good changes. Can we please have a new release?
Last release is now over a year old 😮

pointlessone:
@Zyphrax Is there anything specific you need that is on master? Is there anything preventing you from using master?

Zyphrax:
@pointlessone See my first post.

The release version (2.2.0) causes warnings when your Rails bundle folder is a symlink, which is quite a common scenario for production environments. 

At the moment my Gemfile is pointing to that specific commit, but it slows down bundle updates a lot.

I see a lot of useful changes in the list of commits.
Can I help to get the new version out?

pointlessone:
@Zyphrax. No, not really. A new release will be issued once the work on TTFunk concludes.

jaredbeck:
Looking forward to using pdf-core 0.8. Thanks for all your work on prawn. :heart:

zenspider:
@pointlessone Can you _please_ do a bug fix/point release before you do this bigger release? Getting the dependencies synced up properly and any other bug fixes would be nice before something major comes down the pipe.

Zyphrax:
@pointlessone You would make so many people happy with a small release. It has been almost 18 months now since the last version (Prawn 2.2.0) was released. Like @zenspider said, it would be great to have a small release before any big changes are introduced.

brendandeere:
Would also love a patch release including https://github.com/prawnpdf/prawn/pull/1065

shanecav84:
[...]

PR #1101, opened 2019-02-07
Test on Ruby 2.6 on Travis CI

aried3r:
I saw that this gem is being tested in the lowest and highest patch version of a Ruby release so I did the same with Ruby 2.6. But might I suggest only testing on the very latest patch version? It would greatly reduce test runs and I don't think there is much of a benefit of doing this. WDYT?


pointlessone:
Hi, Anton. Thanks for the contribution.

Historically, there were breaking changes in tiny ("patch" in SemVer vocabulary) releases of Ruby. Generally, builds are ran in parallel. Even though there's some overhead it's far from double. In that regards dropping JRuby would be much significant gain but we're not doing it.

While you're at it, I wonder if you have a bit more time to contribute the same change to PDFCore and TTFunk?

aried3r:
> While you're at it, I wonder if you have a bit more time to contribute the same change to PDFCore and TTFunk?

I'll certainly try! On that note, do you think there'll be a new release of prawn soon?

pointlessone:
There definitely will be a release at some point but probably not "soon" in a conventional meaning. Soon™.

aried3r:
> Soon™

Haha, fair enough. I think there's [quite some people](https://github.com/prawnpdf/prawn/issues/1052) waiting for it. I don't really know much about the internals of prawn, but is there anything you want to do before the next release that's keeping it from being released?

Test pass, so I'd even use a `.pre` version if available on RubyGems :)

pointlessone:
I'm aware about that issue.

I want to update code style. Basically, update Rubocop dep and make it happy. And there's a big feature coming to TTFunk. These two things need to happen for the next release of all three gems. The feature takes some time because it's big and complex. Rubocop though can be done by someone else. I intend to do it myself but life gets priority.

Issue #253, opened 2011-07-11
Problems width_of method


hidakatsuya:
Hi, I'm Japanese.
Sorry, the following text is translated by Google.

First, if a string contains Japanese characters in the case of Ruby1.8, `string.length` does not seem to work correctly to return the number of bytes.

Also, how to calculate the width of space between characters, what do I think that the following is correct?

``` ruby
def width_of(string, options={})
  font.compute_width_of(string, options) + character_spacing * (string.length - 1)
end
```

In the case of Ruby1.8 computes the following characters as you do?

``` ruby
def in_utf8
  original = $KCODE
  result = yield
  $KCODE = original
  result
end

def width_of(string, options={})
  str_length = in_utf8 { string.split(//).size }
  font.compute_width_of(string, options) + character_spacing * (str_length - 1)
end
```

Thank you.

PR #1117, opened 2019-05-13
Fix character_spacing effect on text width calculation.

mtyaka:
The value of `character_spacing` should be added to the length of the original string `(string.length - 1)` times instead of `string.length` times, since character_spacing only increases the space between
characters.

Incorrect text width calculation was causing `:center` and `:right` aligned text to be incorrectly positioned when using `character_spacing != 0` (see screenshots).

Note: this was briefly noticed/discussed in https://github.com/prawnpdf/prawn/issues/253.

**Screenshots - current behavior:**

| Alignment | Screenshot |
| --- | --- |
| left | <kbd></kbd> |
| center | <kbd></kbd> |
| right |<kbd></kbd> |

**Screenshots - with this patch:**

| Alignment | Screenshot |
| --- | --- |
| left | <kbd></kbd> |
| center | <kbd></kbd> |
| right |<kbd></kbd> |

### DIFF
--- a/lib/prawn/font_metric_cache.rb
+++ b/lib/prawn/font_metric_cache.rb
@@ -37,9 +37,12 @@ module Prawn
 
       length = @cache[key]
 
-      length +
-        (@document.character_spacing *
-         @document.font.character_count(encoded_string))
+      character_count = @document.font.character_count(encoded_string)
+      if character_count.positive?
+        length += @document.character_spacing * (character_count - 1)
+      end
+
+      length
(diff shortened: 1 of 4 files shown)


pointlessone:
Thank you for your contribution! Both the PR and the code are well done!

johnnyshields:
Thank you for this!!

mtyaka:
Thanks for the prompt review @pointlessone!

PR #1122, opened 2019-07-25
resolves #1121 read padding from correct box when creating new page

### LINKED ISSUE #1121: Padding is read from wrong box when creating new page
When creating a new page, the padding is read from the margin box instead of the current bounding box. This can cause the indent function to effectively remove the padding twice, leading to a left margin that migrates further and further to the left.

I reported a similar problem in prawn-table. See https://github.com/prawnpdf/prawn-table/issues/114

What it comes down to is that the bounding_box and indent functions don't play well together. Here's a script that reproduces the problem:

```rb
Prawn::Document.generate 'test.pdf' do
  initial_absolute_left = bounds.absolute_left
  text 'paragraph'
  bounding_box [0, cursor], width: bounds.width do
    text 'paragraph in bounding box'
    bounds.move_past_bottom
    text 'paragraph in bounding box'
  end
  indent 20 do
    bounds.move_past_bottom
  end
  if bounds.absolute_left != initial_absolute_left
    warn 'bounds were not properly restored after call to indent'
  end
  text 'paragraph'
end
```

The problematic logic is here: https://github.com/prawnpdf/prawn/blob/c5842a27b15f912f2f0ad5818a9ef38992978b3c/lib/prawn/document.rb#L723-L731

I believe the following logic will resolve the problem:

```rb
# update bounding box if not flowing from the previous page
unless @bounding_box && @bounding_box.parent
  old_margin_box = @bounding_box
  @bounding_box = @margin_box
end

# maintains indentation settings across page breaks
if old_margin_box
  @margin_box.add_left_padding(old_margin_box.total_left_padding)
  @margin_box.add_right_padding(old_margin_box.total_right_padding)
end
```

Not only does it fix the example scenario I provided, the change also passes all tests in Asciidoctor PDF (which uses bounding boxes extensively).

mojavelinux:


mojavelinux:
The failures seem to be due to new cop rules that showed up which are not related to this change.

mojavelinux:
I believe this solution is still correct. It fixes at least 2 different cases Prawn mangles the margins (namely when using bounding_box or column_box).

tomprats:
I'm also running into this issue. Anything I can do to help move this along?

@pointlessone anything you see?

mojavelinux:
I'm willing to help to. I'll will rebase to fix the merge conflict if there is interest.

pointlessone:
A test would be nice to have.

tomprats:
Here's a test based on @mojavelinux's example in the issue https://github.com/tomprats/prawn/commit/e46a4ee8b64ab79a9e68ba7f92389cd1dd4712f4

And here's a branch with @mojavelinux's fix that allows the test to pass (but I didn't cherrypick it so it doesn't attribute the work properly) https://github.com/tomprats/prawn/tree/fix-padding.

@pointlessone - Let me know if you'd like more tests!

@mojavelinux - I can create a new PR if you're busy, or if you just add a similar test case after your rebase, we might be good to go!

mojavelinux:
@tomprats Thanks for the test! I added it to the PR and rebased.

I tried to come up with another test scenario, but I couldn't remember what the situation was that led me to sending this PR. I know it had something to do with indent. Regardless, I can confirm that without the patch, the test does not pass. So I think it's proving the need for the change perfectly.

mojavelinux:
I did a bit more investigation and discovered the scenario where I ran into this problem. However, I can't come up with any other example that demonstrates the problem without involving other libraries such as prawn-table. The issue comes down to the order in which the indentation is applied and the bounding box is restored. The steps need to be in the other order.

pointlessone:
Thank you for your contributions. I will take a look soon.

Issue #1135, opened 2019-11-20
Repeater should be added before any other content (or allow it)


estani:
We use different background images for different pages, adding it with a repeater this way:
```
repeat(->(pg) { pg >= 2 }) do
      blend_mode(:Overlay) do
        load_background_image
      end
    end
```
We need to do it in Overlay mode, because the text is rendered behind the image. The overlay affects the rendering so we want to get rid of it.

I think what is happening is that the repeat block (or stamp) is called after the page has been rendered, not before.

pointlessone:
Repeaters are ran right before the document gets rendered. That is, after all content has been placed. That is by design.

Is there a reason you can not render images before placing the content?

estani:
You mean outside of a repeater? (because as you said, repeaters are called "after all content is placed", so rendering images "before" means not using a repeater, right?)
That will require some callback, after a new page gets rendered. Pretty much like a repeater... hence that's what the PR is about.
Is there another way to get an image pasted depending on the page number "before" the content is placed that I'm not aware of? (perhaps that was your question? In that case the reason I can't do it, is because I don't know how).

To summarize the problem again: A repeater is like a stamp (it is implemented like it). Now you might want to have a stamp behind some text or above, the later is what's implemented, for the former there's the PR.

I'll be glad to make changes if that's required. I did try to follow the code logic as good as I could, but I'm by no means a prawn expert...

pointlessone:
There are a number of solutions.

1. You can use document background:
   ```ruby
   Prawn::Document.generate "test.pdf", background: 'background.png' do
     text 'foo'

     start_new_page
     text 'bar'

     start_new_page
     text 'baz'
   end
   ```
2. You can use `on_page_create` callback:
   ```ruby
   Prawn::Document.generate "test.pdf" do
     on_page_create do
       image 'data/images/prawn.png'
     end

     text 'foo'

     start_new_page
     text 'bar'

     start_new_page
     text 'baz'
   end
   ```

   However note that callback is not called for the first page of the document.

estani:
If I understood your code and comment correctly, this would not work:
1) cannot use different backgrounds for different pages
2) cannot add a background to the first page (from your comment)

From your comments the best option would be to call on_page_create for the first page too (and by the way pass some context information, like the page count)

pointlessone:
It's not quite possible to call `on_page_create` because the first page is created before it can be defined. As a work around you can pain the background on the first page manually.

[...]

PR #1136, opened 2019-11-20
add: renderers run just after new page inserted

estani:
To better discuss #1135


petergoldstein:
@pointlessone Given the discussion in #1135 I don't think we need to keep this PR open anymore, do we?  What do you think - can we close?

estani:
closed as request.

PR #1148, opened 2020-02-23
Look for glyph in correct font when fallback fonts are enabled

### LINKED ISSUE #1147: find_font_for_this_glyph doesn't always look in the correct font
The find_font_for_this_glyph method does not consider the style when resolving the font. As a result, it may be looking in the wrong font and thus return an incorrect result.

The problem is here:

```
fragment_font = hash[:font] || original_font
```

The code only retrieves the font family (`:font`) without taking into account the style. So if the fragment is bold, the code will look for the char in the normal font instead of the bold font.

mojavelinux:
note that this patch adds an optional fourth argument to find_font_for_this_glyph

resolves #1147

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## PrawnPDF master branch
 
+## Unreleased
+
+### Look for glyph in correct font
+
+Take the font style into account when looking for a glyph and fallback fonts are enabled.
+
+(Dan Allen, [#1147](https://github.com/prawnpdf/prawn/issues/1147))
+
 ## PrawnPDF 2.4.0
 
 ### Added support for Ruby 3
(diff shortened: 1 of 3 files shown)


pointlessone:
@mojavelinux Thank you for your contribution.

Could you please rebase the branch on top of master?

mojavelinux:
Sure thing!

mojavelinux:
The PR is now updated.

mojavelinux:
I have updated again.

pointlessone:
@mojavelinux Could you please rebase this PR on master?

mojavelinux:
I have updated again.

pointlessone (approved):
✨

pointlessone:
@mojavelinux Thank you very much.

mojavelinux:
Thanks @pointlessone!

PR #1171, opened 2020-09-01
Update ruby_version requirement to allow ruby 3.0

casperisfine:
Ruby master branch recently bumped the version number to 3.0: https://github.com/ruby/ruby/commit/21c62fb670b1646c5051a46d29081523cd782f11

This is breaking our ruby-head CI.

### DIFF
--- a/prawn.gemspec
+++ b/prawn.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
       '.yardopts'
     ]
   spec.require_path = 'lib'
-  spec.required_ruby_version = '~> 2.5'
+  spec.required_ruby_version = '>= 2.5'
   spec.required_rubygems_version = '>= 1.3.6'
 
   spec.authors = [


pointlessone:
Could you please rebase this PR on master?

casperisfine:
Done.

pointlessone:
@casperisfine Is there a docker container for ruby-head we could use to run our tests in CI?

casperisfine:
Yes, there is https://github.com/ruby/ruby-docker-images which builds nightly images. But looking at your CI, I think you could simply use the default `setup-ruby` action. See how psych does it for instance: https://github.com/ruby/psych/blob/181a727c90cb1d320db5821bc16dc0bc34757935/.github/workflows/ubuntu.yml

casperisfine:
Do you want me to PR a CI revamp?

pointlessone:
I definitely wouldn't mind that. :)

Here are a few things to keep in mind. A wishlist of sorts.

* We want to keep all our current rubies. That includes JRuby. For CRuby we have the first release (*.*.0) and the latest in the series (e.g. 2.7.1 for 2.7).
* We'd love to be able to add new rubies without a whole CI revamp. There are people who'd like to add TruffleRuby, for example.
* We'd like to have separate builds for Release Candidates and Head.
* We might be interested in adding JRuby-head.

The first point is sort of a given. There rest are of an informational nature just to give you a broader context.

pointlessone (approved):

PR #1173, opened 2020-09-05
Update ci.yml

gogainda:


.github/workflows/ci.yml
```diff
@@ -10,23 +10,17 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-    container: ${{matrix.image}}
     strategy:
       fail-fast: false
       matrix:
-        image:
-          - ruby:2.5.0
```
pointlessone, line 18: I appreciate your enthusiasm about TruffleRuby but I'd like to keep our current set of rubies as well.

gogainda: I will revert it back, just want to skip all other ruvies for now

pointlessone: OK. Let me know when this PR is ready.

gogainda:
@pointlessone  tests are passing, but rubocop fails due to a [bug](https://github.com/oracle/truffleruby/issues/1913), so without upgrading rubocop to the 0.90 truffleruby cannot be enabled right now

gogainda:
I tried to upgrade the rubocop version to 0.90 and it helped, but it introduced some failing checks which are part of newer version of rubocop.

eregon:
@gogainda Could you just skip RuboCop on TruffleRuby? It would be redundant anyway and just make the build slower.

pointlessone:
I tried adding truffleruby in the latest CI update and encountered some issues.

Specifically, build fails with truffleruby and the failure is some sort of error in Rubocop.

At the moment I don't see truffleruby feasible. Especially since there's no one on the team using it. We'll gladly accept help from someone who's willing to spend their time regularly fixing compatibility issues.

eregon:
Makes sense, I'll make a draft PR and try to get it working.
It would be valuable to test latest Prawn + TruffleRuby since it's a non-trivial codebase and actually all tests pass fine.

eregon:
https://github.com/prawnpdf/prawn/pull/1189

PR #1175, opened 2020-09-07
Refactor CI to test against ruby-head

casperisfine:
As discussed in https://github.com/prawnpdf/prawn/pull/1171

This simplify the CI by using the standard setup ruby action.

ruby-head fails for now, but https://github.com/prawnpdf/prawn/pull/1171 will fix it.

cc @pointlessone

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,27 +9,30 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-    container: ${{matrix.image}}
     strategy:
       fail-fast: false
       matrix:
-        image:
-          - ruby:2.5.0
-          - ruby:2.5
-          - ruby:2.6.0
-          - ruby:2.6
-          - ruby:2.7.0
-          - ruby:2.7
-          - jruby:9.2
-        include:
-          - image: ruby:rc
-            continue-on-error: true
+        ruby:
+          - 2.5.0
+          - 2.5
+          - 2.6.0
+          - 2.6
+          - 2.7.0
+          - 2.7
+          - ruby-head
+          - jruby-9.2
+          - jruby-head
+
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
+
       - name: Set up Ruby
-        run: |
-          gem update --system > /dev/null
-          gem --version
-          gem install -g
-      - name: Run tests
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - name: Install dependencies
+        run: bundle install --jobs 4 --retry 3
+
+      - name: Run test
         run: rake


casperisfine:
Sorry, just saw your last comment. I'll address it.

casperisfine:
> We'd like to have separate builds for Release Candidates and Head.

I'm not so sure what you mean by this. 

The other points were addressed. Note that right now it run both this branch and master actions, but once merged there won't be as many duplicated statuses.

.github/workflows/ci.yml
```diff
@@ -2,34 +2,37 @@ name: CI
 
 on:
   push:
-  pull_request_target:
+  pull_request:
```
pointlessone, line 5: Why this change?

.github/workflows/ci.yml
```diff
@@ -2,34 +2,37 @@ name: CI
...
-          gem update --system > /dev/null
-          gem --version
-          gem install -g
-      - name: Run tests
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - name: Install dependencies
+        run: bundle install --jobs 4 --retry 3
```
pointlessone, line 35: Is bundler available with all rubies?

.github/workflows/ci.yml
casperisfine: Because it wouldn't run my change on this PR, and I couldn't find any reference to `pull_request_target ` in the GitHub Actions documentation.

.github/workflows/ci.yml
```diff
@@ -2,34 +2,37 @@ name: CI
...
-          - ruby:2.6.0
-          - ruby:2.6
-          - ruby:2.7.0
-          - ruby:2.7
-          - jruby:9.2
-        include:
-          - image: ruby:rc
-            continue-on-error: true
+        ruby:
+          - 2.5
```
pointlessone, line 16: My understanding is this means "the latest in 2.5 line". We specifically want this and 2.5.0. Is that possible?

.github/workflows/ci.yml
```diff
@@ -2,34 +2,37 @@ name: CI
...
+          - 2.6
+          - 2.7
+          - ruby-head
```
pointlessone, line 19: We don't want to break the build on unstable rubies failures. These are only so that we knew we have some work before we can support the next release. There's also a chance that unstable rubies might have bugs.

.github/workflows/ci.yml
```diff
@@ -2,34 +2,37 @@ name: CI
...
+          - jruby-9.2
+          - jruby-head
     steps:
-      - uses: actions/checkout@v1
+
+      - name: Install libraries
+        run: sudo apt install haveged
```
pointlessone, line 25: I don't think we need this.

.github/workflows/ci.yml
casperisfine: Yes.

.github/workflows/ci.yml
casperisfine: Yes it is:

```
- 2.5.0
- 2.5
```

.github/workflows/ci.yml
casperisfine: Unfortunately that feature doesn't really exist on GitHub Actions.

I suppose I could move the RCs to another pipeline, and have a || true, but then it will show as green on GitHub so the value is limited.

.github/workflows/ci.yml
casperisfine: Yes, total brainfart...

.github/workflows/ci.yml
casperisfine: Ok, that explains it then. I'll leave the change for now otherwise it's hard for me to test my changes. But I can change it back last minute.

.github/workflows/ci.yml
pointlessone: It does exist. It's not obvious but here's how it's done. You include a duplicate entry in the matrix and add `continue-on-error: true` to it. See Line 24 in the current config.

casperisfine:
Should be good now. I did split the -head version and set continue-on-failure for them (only on the `rake` command, hence why `ruby-head` still fail as it doesn't pass bundle install)

pointlessone:
I'd rather we have all CI builds in one file. Especially since they're conceptually the same. Allow the whole build to fail for pre-release rubies as we do now.

casperisfine:
Done.

.github/workflows/ci.yml
```diff
@@ -2,34 +2,58 @@ name: CI
...
-          - image: ruby:rc
-            continue-on-error: true
+        ruby:
+          - 2.5.0
+          - 2.5
+          - 2.6.0
+          - 2.6
+          - 2.7.0
+          - 2.7
+          - jruby-9.2
```
pointlessone, line 22:
```suggestion
          - jruby-9.2
        include:
          - ruby: ruby-head
            continue-on-error: true
          - ruby: jruby-head
            continue-on-error: true
```

This way we don't need `test-rc` section.

casperisfine: That doesn't work. See the last commit I pushed.

casperisfine:
Ok, I made it work by making that a parameter of the matrix.

.github/workflows/ci.yml
pointlessone: I'm not quite sure what you mean? How can you tell it doesn't work?

casperisfine: It was still failing. `continue-on-error` can only be set on a step, not on a job. I worked around that with an interpolation.

pointlessone: I was under impression that this would be set on a job (in our case `test`), not a step. I got that from [the documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error). But I'm not quite sure how it interacts with the matrix.

casperisfine: Well, I pushed another change with that, and unless I'm overlooking something it doesn't work.

pointlessone: I'm getting confused. What does `continue-on-error` do? One of the previous runs just marked "head" build green even though it failed.

That's one way to interpret it but it hides the failure. We already have set up a couple build that must be green in order for PR to be accepted (see "Required" label in the checks list). I'd still want to know when particular builds fail.

Maybe we don't need this `continue-on-error` at all? I certainly don't want to dig through all build to find which are failing. I'd rather see them red in the checks list.

casperisfine: > What does continue-on-error do? One of the previous runs just marked "head" build green even though it failed.

Yes that's exactly what it does, and that is why it's a very bad replacement for Travis's `allow_failures` feature. It show the step as green, so you might break the build and nobody will ever notice.

> Maybe we don't need this continue-on-error at all? I certainly don't want to dig through all build to find which are failing. I'd rather see them red in the checks list.

That is exactly what I was initially suggesting.

pointlessone: All right. Sorry for the confusion.

I must've mixed it up `fail-fast`.

Let's let particular checks fail but run all of them regardless if any one fails.

casperisfine: Done. I also restored the `pull_request_target`, should be good to merge now.

pointlessone (approved):

pointlessone:

I reworked CI workflow largely based on this PR.

I haven't merged your PR because I had to change it a bit and test it a bunch to make it suitable for other Prawn gems too.

PR #1176, opened 2020-09-07
replace require_relative in spec helper

boutil:
Hi,
according to the following document,
https://github.com/rubocop-hq/packaging-style-guide#using-require-relative-from-test-to-lib
it is considered a good practive to not use require_relative to code in lib from the spec.
This would help running the tests of the gem in distributions providing packaged versions of prawn, like Debian.
Thanks in advance for considering this change.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -9,7 +9,7 @@ if ENV['COVERAGE']
   end
 end
 
-require_relative '../lib/prawn'
+require 'prawn'
 
 Prawn.debug = true
 Prawn::Fonts::AFM.hide_m17n_warning = true


pointlessone:
First of all, thank you for your contribution.

I'm slightly skeptical about that particular packaging guide. It's written by Debian maintainers and Debian is known to have very specific packaging requirement but they're not necessarily universally applicable. They have their own testing and packaging infrastructure which is not RubyGems and is not reflective of the wider gems usage.

I will get back to this MR later when I'll have some time to reflect on it's content.

pointlessone:
I'll close this MR. I'm not convince the guide is reflective of the generally accepted practice in the Ruby community.

You're invited to participate in the further discussion: #1181.

PR #1186, opened 2020-12-09
Add Truffleruby head to CI

gogainda:


pointlessone:
I tried adding truffleruby in the latest CI update and encountered some issues.

Specifically, build fails with truffleruby and the failure is some sort of error in Rubocop.

At the moment I don't see truffleruby feasible. Especially since there's no one on the team using it. We'll gladly accept help from someone who's willing to spend their time regularly fixing compatibility issues.

PR #1187, opened 2020-12-23
Fix typo in documentation for bounding box nesting

clinejj:
There was a misspelling of `outer`, so small PR to correct.


pointlessone:

PR #1189, opened 2020-12-30
Add TruffleRuby in CI

eregon:
I marked as a Draft PR because there is some spec failing with OutOfMemoryError on TruffleRuby.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
           - "3.0.0"
           - ruby-head
           - jruby-9.2
+          - truffleruby-head
     steps:
       - uses: actions/checkout@v1
       - name: Set up Ruby
--- a/spec/prawn_manual_spec.rb
+++ b/spec/prawn_manual_spec.rb
@@ -5,7 +5,7 @@ require 'digest/sha2'
 
 MANUAL_HASH =
   case RUBY_ENGINE
-  when 'ruby'
+  when 'ruby', 'truffleruby'
     'de26db4fe63e024231c0a332203b41305103d877b584a2e98dbd0561bced39f2c066b5c0c96a4686e586a9deb347f099dac4c646446dadb1521a7d4a674ae6fb'
   when 'jruby'
     'c002ffaf6fe4b2877bd2244735e99c04a4b28b06bc365f343411af052d491660e0d858a956a757ad15a4ed16d6808fc8d726fd683d524f5a3f7c0c8b9566b683'


pointlessone:
@eregon Will you be interested in supporting Prawn on TruffleRuby?

eregon:
Yes, as in I'd check the CI (via https://github.com/eregon/truffleruby-gem-tracker) and if it fails I'd take a look at it and fixing whatever issue there is.

Rakefile
```diff
@@ -3,7 +3,8 @@
 GEMSPEC = File.expand_path('prawn.gemspec', __dir__)
 require 'prawn/dev/tasks'
 
-task default: %i[spec rubocop]
+task default: :spec
+task default: :rubocop if RUBY_ENGINE == 'ruby'
```
pointlessone, line 7: I don't this is needed any more.

eregon:
Rebased, and I removed the commits related to #1190.

Rakefile
eregon: It seems wasteful to run RuboCop on TruffleRuby, and there might be issues with RuboCop on TruffleRuby, and those are just not interesting to check in this CI.

pointlessone: We don't in CI and locally you have a choice to run Rubocop and RSpec separately.

eregon: Ah, but the CI does `bundle exec rake spec` now, I see, I'll remove the commit.

pointlessone: Yes, it's a recent change.

.github/workflows/ci.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           - "3.0.0"
           - ruby-head
           - jruby-9.2
+          - truffleruby-head
```
pointlessone, line 48: I'm not seeing a job for TruffleRuby in the latest builds. Any idea what might cause that?

eregon: It seems due to the workflow using `pull_request_target:` instead of `pull_request:`

eregon: @pointlessone  Any reason why `pull_request_target` is used?
Could you use the more intuitive pull_request, as documented [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-list-of-events) ?
Note that changing that in a PR would likely not work, due to how `pull_request_target` seem to ignore workflow changes in PRs.

pointlessone: When we migrated to GH Actions PR from forks wouldn't trigger CI builds. We've settled on the current setup but I'm not sure if it's entirely correct. I'll read the docs again and see if it needs changing.

Meanwhile, I checked builds on your fork and see that a few specs failed with OOM. I tried locally and for a few specs memory usage goes north of 29GB and the whole test suite takes over 10 min on 400%+ CPU while MRI manages run it on a single core in 13s and a few hundreds MB of RAM. Is this expected or is my TruffleRuby not quite right? I've got mine from rbenv, it's `truffleruby 21.1.0-dev-70c024bb`.

eregon: > When we migrated to GH Actions PR from forks wouldn't trigger CI builds. We've settled on the current setup but I'm not sure if it's entirely correct. I'll read the docs again and see if it needs changing.

I know that `pull_request` works just fine, and clearly `pull_request_target` doesn't work here.
Yes, `pull_request` needs to be in `on:` explicitly.

> Meanwhile, I checked builds on your fork and see that a few specs failed with OOM. I tried locally and for a few specs memory usage goes north of 29GB and the whole test suite takes over 10 min on 400%+ CPU while MRI manages run it on a single core in 13s and a few hundreds MB of RAM. Is this expected or is my TruffleRuby not quite right? I've got mine from rbenv, it's `truffleruby 21.1.0-dev-70c024bb`.

Something is definitely not right. There is an upcoming fix related to memory consumption for many String concatenations,  that should be merged soon. We'll take a look after that's merged.
29GB seems like a pathological case, the GitHub runners have 7GB of RAM, and they seem to OOM reliably on a single test (`Prawn::Images::PNG When making a pdf file with png images and the image is Type 6 does not error`), so I guess that will be the interesting test to look at.
More CPU usage (in %) is expected, especially for test suites, as the JIT uses those other cores.

pointlessone: >  pull_request needs to be in on: explicitly.

Thank you. I'll look into it.

> There is an upcoming fix related to memory consumption for many String concatenations, that should be merged soon.

Do you mind postponing merging this PR until after it lands?

eregon: > Do you mind postponing merging this PR until after it lands?

Yes of course, the new CI job needs to pas before merging :)

pointlessone: Please ping me when that happens.

eregon: I think the OOM error above was related to https://github.com/oracle/truffleruby/issues/2127, which is now fixed.

Could you switch from `pull_request_target` to `pull_request` so CI runs on PRs?

PR #1193, opened 2021-01-08
Add valign validation

masakinpo:
valign for box should have some validation as align does

If there is any reason to have no restriction on valign, please let me know!


pointlessone (approved):

PR #1214, opened 2021-07-21
Embedded file support for Prawn documents

dvdalilue:
Depends on pdf-core PR [47](https://github.com/prawnpdf/pdf-core/pull/47)

This PR defines a new module that allows attaching files (or data) into the generated documents. The module relies on two classes, each one describes a pdf object needed to embed files, the `Filespec` and `EmbeddedFile` objects.

The tests and documentation are missing but I'm working on the second one while updating the pdf-core PR [47](https://github.com/prawnpdf/pdf-core/pull/47) with some tests. Then I'll focus on this PR more deeply.


lib/prawn/embedded_files/filespec.rb
```diff
@@ -0,0 +1,42 @@
...
+        @file_name = PDF::Core::LiteralString.new(name)
+
+        if options[:description]
+          desc_str = options[:description].to_s
+          @description = PDF::Core::LiteralString.new(desc_str)
+        end
+
+        if options[:hidden].nil? || options[:hidden] != true
+          @not_hidden = true
+        end
```
gettalong, line 20: Can't you just say `@not_hidden = !@options[:hidden]`?

Also: The only reference for this is on line 28 where the negation is wanted. Why not just store `@hidden` directly and rename the method to `#hidden?`? Then one would not need to mentally do "not not hidden".

lib/prawn/embedded_files.rb
```diff
@@ -0,0 +1,43 @@
...
+    # Link to PR (https://github.com/prawnpdf/pdf-core/pull/47)
+
+    def file(src, options = {})
+      path = Pathname.new(src)
+
+      if path.directory?
+        raise ArgumentError, 'Data source can\'t be a directory'
+      elsif path.file?
+        data = path.read
+        options[:name] ||= src
```
gettalong, line 15: You could add the options `:modification_date` and `:creation_date` here based on `path.mtime` and `path.ctime`.

dvdalilue: Sorry, I missed that. Fixed.

lib/prawn/embedded_files/filespec.rb
```diff
@@ -0,0 +1,39 @@
...
+        name = options[:name] || file.chksum
+
+        @file_name = PDF::Core::LiteralString.new(name)
+
+        if options[:description]
+          desc_str = options[:description].to_s
+          @description = PDF::Core::LiteralString.new(desc_str)
+        end
+
+        @hidden = !options[:hidden].nil? && options[:hidden] == true
```
dvdalilue, line 18: @gettalong I simplified the expression to avoid the "not not". Although, I wouldn't like to use just `options[:hidden]` as a condition because any object (e.g. `'false'`) will be `true`. So I thought it would be more restrictive to require specifically `true` as the option's value

lib/prawn/embedded_files/filespec.rb
dvdalilue: Yes and no, I was overthinking about this a little bit. I have made some changes and left a comment below with a brief explanation

lib/prawn/embedded_files/filespec.rb
dvdalilue: Checking for `!options[:hidden].nil?` is avoidable, so I removed too

lib/prawn/embedded_files/embedded_file.rb
```diff
@@ -0,0 +1,42 @@
...
+require 'digest/md5'
+
+module Prawn
+  module EmbeddedFiles
+    class EmbeddedFile
+      attr_reader :chksum
+
+      def initialize(data, options = {})
+        @creation_date = options[:creation_date]
+        unless @creation_date.instance_of?(Time)
```
gettalong, line 12: Why not `kind_of?(Time)` here and on line 17?

lib/prawn/embedded_files/embedded_file.rb
```diff
@@ -0,0 +1,42 @@
...
+          @creation_date = Time.now.utc
+        end
+
+        @mod_date = options[:modification_date]
+        unless @mod_date.instance_of?(Time)
+          @mod_date = Time.now.utc
+        end
+
+        @chksum = Digest::MD5.hexdigest(data)
```
gettalong, line 21: You are using `#hexdigest` here where you should be using `#digest`. Otherwise the value isn't the expected 16 byte digest string.

lib/prawn/embedded_files/embedded_file.rb
```diff
@@ -0,0 +1,42 @@
...
+        @data = data.dup
```
gettalong, line 22: Duplicating the data here seems wasteful - any particular reason? Especially, since it is just appended to the stream later on in line 36.

lib/prawn/embedded_files/filespec.rb
gettalong: Any value other than `nil` and `false` are `true`. So if you pass in a string, it is a truthy value. Generally, if you mis-use an API you need to expect wrong outcomes. Therefore I wouldn't check for `== true` but for a truthy value.

dvdalilue: Ok, sounds fair

lib/prawn/embedded_files/embedded_file.rb
dvdalilue: Sure, `kind_of?` seems more appropiate

lib/prawn/embedded_files/embedded_file.rb
dvdalilue: Ok, thanks for the clarification

lib/prawn/embedded_files/embedded_file.rb
dvdalilue: Indeed, I was a little hesitant about duplicating or not. I was thinking about hypothetical cases that may never occur. Fixed

gettalong:
Okay, I think there is nothing to do here for the time being without the rest of the functionality as per the comments.

lib/prawn/embedded_files.rb
```diff
@@ -0,0 +1,45 @@
...
```
pointlessone, line 15: We probably should not mutate passed in arguments.

lib/prawn/embedded_files.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Prawn
+  module EmbeddedFiles
+    # include PDF::Core::EmbeddedFiles
+    # Link to PR (https://github.com/prawnpdf/pdf-core/pull/47)
+
+    def file(src, options = {})
```
pointlessone, line 8: Please add documentation for this method.

lib/prawn/embedded_files/embedded_file.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'digest/md5'
+
+module Prawn
+  module EmbeddedFiles
+    class EmbeddedFile
+      attr_reader :chksum
```
pointlessone, line 8: Is there a reason not to spell it out completely?

lib/prawn/embedded_files/filespec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      def build_pdf_object(document)
+        obj = document.ref!(
+          Type: :Filespec,
+          F: @file_name,
+          EF: { F: @file },
+          UF: @file_name
+        )
+
+        obj[:Desc] = @description if @description
```
pointlessone, line 34: I believe this would fail. `obj` here is a reference and it doesn't seem to have a `[]=` method. We probably want `obj.data[:Desc] = ...` here.

pointlessone, line 34: On a separate note, please, don't use ivars outside of `initialize` or trivial accessors. Add a private `attr_reader` and use that instead.

lib/prawn/embedded_files/embedded_file.rb
```diff
@@ -0,0 +1,42 @@
...
+          Params: {
+            CreationDate: @creation_date,
+            ModDate: @mod_date,
+            CheckSum: PDF::Core::LiteralString.new(@chksum),
+            Size: @data.length
+          }
+        )
+
+        obj << @data
+        obj.stream.compress!
```
pointlessone, line 37: Document has a global option that controls compression so we should honour it here. This should be behind `if document.compression_enabled?` guard.

pointlessone (changes requested):
I would love some tests.

pointlessone:
Thank you for your contribution. This is a great start.

lib/prawn/embedded_files/filespec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      def hidden?
+        @hidden
+      end
+
+      def build_pdf_object(document)
+        obj = document.ref!(
+          Type: :Filespec,
+          F: file_name,
+          EF: { F: file },
```
gettalong, line 30: Since the file specification dictionary contains the /F and /UF entries, it would makes sense to also use the /UF entry inside the /EF key, i.e. `EF: { F: file, UF: file }`.

I don't think this is required but it doesn't add much in terms of PDF size and we would be on the safe side if some PDF readers expect the key to be there.

dvdalilue: Sure, probably it's better to use it

lib/prawn/embedded_files.rb
dvdalilue: Done

PR #1225, opened 2021-11-14
Add Font#font_style and Font#font_family

c960657:
Just like you can use `font_size` to adjust the size without passing the current font name, add `font_style` to allow changing the style while preserving the font size and family.

### DIFF
--- a/lib/prawn/font.rb
+++ b/lib/prawn/font.rb
@@ -47,12 +47,15 @@ module Prawn
     # font files.
     # See font_families for more information.
     #
-    def font(name = nil, options = DEFAULT_OPTS)
-      return((defined?(@font) && @font) || font('Helvetica')) if name.nil?
+    def font(name = nil, options = nil)
+      return @font || font('Helvetica') if name.nil? && options.nil?
 
       if state.pages.empty? && !state.page.in_stamp_stream?
         raise Prawn::Errors::NotOnPage
       end
+
+      name ||= font_family
+      options ||= DEFAULT_OPTS
 
       new_font = find_font(name.to_s, options)
 
@@ -66,6 +69,51 @@ module Prawn
       end
 
       @font
+    end
+
+    def font_family
+      @font ? @font.family : 'Helvetica'
+    end
+
+    # @method font_style(points=nil)
+    #
+    # When called with no argument, returns the current font style.
+    #
+    # When called with a single argument but no block, sets the current font
+    # style.  When a block is used, the font style is applied transactionally and
+    # is rolled back when the block exits.  You may still change the font style
+    # within a transactional block for individual text segments, or nested calls
+    # to font_style.
+    #
+    #   Prawn::Document.generate("font_style.pdf") do
+    #     font_style :bold
+    #     text "Bold text"
+    #
+    #     font_style(:italic) do
+    #       text "Italic text"
+    #       text "Bold text", :style => :bold
+    #       text "Italic text"
+    #     end
+    #
+    #     text "Bold text"
+    #   end
+    #
+    # Font styles are not additive, i.e. if the current style is :bold, setting
+    # the style to :italic results in italic, not bold-italic text. Use
+    # :bold_italic instead.
+    #
+    # When called without an argument, this method returns the current font
+    # style.
+    #
+    def font_style(style = nil, &block)
+      return @font ? @font.style : :normal if style.nil?
+
+      font(font.name, style: style, &block)
+    end
+
+    # Sets the font style
+    def font_style=(style = nil)
+      font_style(style)
     end
 
     # @method font_size(points=nil)
@@ -299,6 +347,9 @@ module Prawn
     # The current font family
     attr_reader :family
 
+    # The current font style
+    attr_reader :style
+
     # The options hash used to initialize the font
     attr_reader :options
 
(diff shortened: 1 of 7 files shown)


manual/text/font_style.rb
```diff
@@ -20,6 +25,11 @@
     styles.each do |style|
       font example_font, style: style
       text "I'm writing in #{example_font} (#{style})"
+
+      font_style style
+      text "This is also #{style}"
+
+      text "And this is also #{style}", style: style
```
gettalong, line 32: I get what you want here but since setting the style using `#font_style` or inline is a no-op, I don't think it's very illustrative. Maybe use the block form of `#font` and `#font_style` to make it clearer?

spec/prawn/font_spec.rb
```diff
@@ -111,11 +111,109 @@
...
+    end
+  end
+
+  describe '#font_style' do
+    it 'returns default font style' do
+      expect(pdf.font_style).to eq(:normal)
+    end
+
+    it 'allows setting font style in DSL style' do
+      expect(pdf.font_style).to eq(:normal)
```
gettalong, line 176: This is already tested in 'returns default font style', so should be removed here.

lib/prawn/font.rb
```diff
@@ -68,6 +71,51 @@ def font(name = nil, options = DEFAULT_OPTS)
...
+    # the style to :italic results in italic, not bold-italic text. Use
+    # :bold_italic instead.
+    #
+    # When called without an argument, this method returns the current font
+    # style.
+    #
+    def font_style(style = nil, &block)
+      return @font ? @font.style : :normal if style.nil?
+
+      font(font.name, style: style, &block)
```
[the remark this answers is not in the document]
gettalong: Reading the associated tests I think the tests might work because `name` and `family` are the same for 'Helvetica'. Could you add a test that uses a registered TrueType font?

gettalong (changes requested):
Thanks for this PR!

I think that this is a feature that can be added without problems once the few questions are addressed.

lib/prawn/font.rb
```diff
@@ -68,6 +71,48 @@ def font(name = nil, options = DEFAULT_OPTS)
...
+    #   end
+    #
+    # Font styles are not additive, i.e. if the current style is :bold, setting
+    # the style to :italic results in italic, not bold-italic text. Use
+    # :bold_italic instead.
+    #
+    def font_style(style = nil, &block)
+      return @font ? @font.style : :normal if style.nil?
+
+      font(font_family, style: style, &block)
```
gettalong, line 108: So, thinking a bit more about this, why not just pass `nil` here? Then it would automatically use what's defined in `#font` and we would have only one place where it would be defined.

gettalong: Another thing that came to mind: What happens if we pass the path to a e.g. TrueType font into `#font` and then call `font_style(:bold)`?

Since nothing is defined and we cannot know which file contains the bold version, I guess the only thing to do is to error out here. What do you think?

spec/prawn/font_spec.rb
```diff
@@ -111,11 +111,123 @@
...
+        expect(pdf.font.name).to eq('Helvetica-Bold')
+      end
+
+      expect(pdf.font_style).to eq(:normal)
+      expect(pdf.font.name).to eq('Helvetica')
+    end
+
+    it 'allows setting font style for a TTF font' do
+      pdf.font_families.update(
+        'DejaVu Sans' => {
```
gettalong, line 193: Thanks for the test case. However, I think to really test this out we need another entry where the key is different from the font family name, e.g. using 'deja' as key.

I don't think that the current implementation would work since the look-up key would be the family name from the TTF.

lib/prawn/font.rb
gettalong: Maybe we should check here for `font_families.key?(font_family)` and error out if nothing is found. This would solve both problems, the one when using a TrueType font path and the one where the used key is different from the family name. The error message should probably include the reason for the failure, i.e. that the family name was not registered.

And updating the documentation for `#font_families` in this regard would also make sense.

gettalong (changes requested):

lib/prawn/font.rb
c960657: I agree that calling `font_style` with a key not present in `font_families` for the current font should raise an error. Likewise if style is other than `normal` for a specific font file.

Currently `font` does not raise in the latter situation, and changing this would be a BC break, so I suggest we avoid raising in that specific situation (`font` called with style other than normal for a specific font file). I imagine that it is not uncommon that people do e.g. `font 'ArialBlack.ttf', style: :bold` in some (mis-guided) attempt to signal that this is a bold font.

gettalong: Agreed on both accounts.

@pointlessone What do you think?

pointlessone: It's a little confusing because we have two very similar things:
- `font_families` which afford for styles
- `font_registry` which does not quite do that.

`font_families` eventually gets mapped to entries in `font_registry`. Style is only meaningful on the stage of this mapping. Whenever a font file is used directly style has no meaning. It still will be present in the font but there's no way to look up a font for any other style. We never add these fonts to `font_families`, we never look up fonts by their internal names so we're unable to correlate fonts from the same family. I don't think this should change now. Maybe in the next major release if we ever get there.

I suggest, we make `font_style(:random_style) { ... }` a noop but issue a `warn`ing. Possibly, when `$DEBUG` is set but it rarely is so it's not very useful.

PR #1227, opened 2021-11-30
Test against Ruby 2.4

hallelujah:
Add Ruby 2.4 to the test matrix so we are sure to be able to deploy it


hallelujah:
Sorry wrong repo

Issue #1195, opened 2021-01-21
Release a version


tatosjb:
Recently was integrated a PR who solves the font cache issue. 

Is possible to release a version? The last release is from December 31.

https://rubygems.org/gems/prawn

donv:
A new release is required for proper dependencies for Ruby 3.1:

https://github.com/prawnpdf/prawn/commit/3658d5125c3b20eb11484c3b039ca6b89dc7d1b7

AlfonsoUceda:
Hi folks, any news on this?

pointlessone:
OK, I will push a new release before the snow melts away. But to set expectations right it won't happen in the next two weeks.

alec-c4:
Hey, folks! Any news on this issue?

rubyFeedback:
A prawn version 3.0 would be pretty neat eventually.

We have some competition e. g. hexapdf by gettalong. I have quite a lot
of pdf-generation code in different projects, so I am re-evaluating old 
code every now and then. One advantage gettalong has is that he is
quite active (and home boy in my city too! \o/ ).

deanpcmad:
Any news on this?

gettalong:
@pointlessone If you need help with a new release, let me know! But no stress, when you have time, you have time..

pointlessone:
@gettalong We've got a bunch of PRs open. I'd love some help with reviews. That extends to the broader community, too. Thank you.

johnnyshields:
@pointlessone any reason the master branch can't be released "as-is"? Are there any specific open PRs blocking the release?

pointlessone:
I don't think any of the PRs are blocking release. It's just an area where I could use some help.

johnnyshields:
Perhaps we should decouple the two? Do a release, and then the good folks on this thread can chip in to help with PRs?

pointlessone:
To be clear, I'm not waiting on PRs to merge before cutting a release. Prawn release is a somewhat complicated process so I need a solid block of time for it. That's the blocker. While we're waiting on that PRs might as well get reviewed.

gettalong:
@pointlessone I have created pull requests https://github.com/prawnpdf/prawn/pull/1242 and https://github.com/prawnpdf/pdf-core/pull/48 for open issues. Those are rather small changes and would be good candidates for merging before making the new release.

zavan:
Is using the master branch directly ok?

pointlessone:
[...]

PR #1234, opened 2022-01-04
Add Ruby 3.1.0 to CI

Kevinrob:
Since [Ruby 3.1.0 is out](https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/), we should add it to the CI workflow.

@pointlessone Can you release a new version? `prawn` 2.4.0 fail with Ruby 3.1.0 without adding `matrix` gem to the Gemfile.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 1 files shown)


pointlessone (approved):

pointlessone:

A new release will come eventually. See https://github.com/prawnpdf/prawn/issues/1195#issuecomment-1002559971

PR #1246, opened 2022-02-09
Add TruffleRuby head to CI

petergoldstein:
This is a clean version of #1189 against current master head.  Thanks to @eregon for the original submission.

I have two concerns with this PR:

1. Truffleruby appears to produce a different hash than either MRI ruby or JRuby for the manual
2. Truffleruby takes ~7 minutes to run specs that take ~10 seconds in MRI.  That seems excessive.

Posting this PR for discussion.  @eregon @pointlessone @gettalong any thoughts are welcome.


johnnyshields (approved):

pointlessone:
I would rather not add TruffleRuby unless I have commitment from someone to support it. TruffleRuby is an exciting development but I don't know if it has a solid userbase to warrant the effort. I don't know anyone who uses it in production and the CI overhead is substantial.

petergoldstein:
@pointlessone I think that's reasonable. @eregon are you interested in signing up for that?  Do you have any idea why the tests are so slow for Truffleruby?

pointlessone:
1. GitHub Actions actions is not great for optional build steps. We can't have a green build with some steps failing. It makes the whole contribution UX much worse.
2. Waiting for the slow non-essential step is suboptimal. GH is adamant about waiting for the build completion to switch PR UI.

In short, adding TruffleRuby doesn't benefit Prawn much.

eregon:
If the CI time is a concern (it's 6min30s in [this run](https://github.com/prawnpdf/prawn/runs/5137930247?check_suite_focus=true), still sounds fairly fast):
* We could try to make it faster by investigating and e.g. using `--engine.Mode=latency`
* It could be a daily or weekly job instead, so then it doesn't affect CI time for PRs

petergoldstein:
@eregon The big concern is the comparative time.  Note that the truffleruby "Run tests" section takes over 5 minutes to run.  As compared to ~15s for each of the MRI rubies.  That's especially odd given that one of Truffle's main selling points is speed.

The scheduled job is a good suggestion though.

eregon:
It's expected more optimizing Ruby JITs run test suites slower (e.g., similar on JRuby), because tests constantly run different code and so rarely run any existing & JIT-compiled code. And the interpreter for a Ruby with a JIT needs to collect profiling data which costs some time, as well as the JIT consuming CPU time.
There are ways to improve this, including persistent the JITed code, and using `TRUFFLERUBYOPT=--engine.Mode=latency` which optimizes startup/warmup over peak performance.
The slowdown shouldn't be that big though, and @bjfish is currently looking into that.

I think the scheduled job is a good solution, especially for `truffleruby-head`, it doesn't get in the way but still regularly produces CI results and we can easily find out if it got red and regressed somehow.

petergoldstein:
@eregon So I can understand that, but as you note it really shouldn't be that big.  A factor of 20 is pretty substantial.  Would be interested to hear the cause.  I've been unable to get truffleruby running locally, so I haven't been able to narrow down the slowdown.

@pointlessone If it sounds good to you, I can change this PR to create a scheduled CI job (daliy?  weekly?) that includes `truffleruby-head`.  Let me know.

eregon:
From the [flamegraph](http://eregon.me/flamegraphs/prawn-flamegraph_20220214-102939.svg) it seems most of time (31.2% using `Search`) is spent in `Prawn::Images::PNG#split_image_data` and specifically in `StringIO#read` and `String#byteslice` called from there. The main cost is probably scanning the substring to compute the correct coderange (7-bit or valid for BINARY strings).
This should improve when TruffleRuby has adopted TruffleString (WIP).
It also indicates that in this case it might be better to compute the coderange lazily, since it's one of the rare cases where the coderange doesn't seem needed (for `color.write data.read(color_bytes)`).
From a wider perspective it's an unfortunate case of Ruby not having a real "byte array" structure, and String being a relatively poor fit for it.
Let's track this as a TruffleRuby issue: https://github.com/oracle/truffleruby/issues/2599

pointlessone:
@petergoldstein We already have a weekly schedule for CI. I'm fine including TruffleRuby in it. I'd skip it in PR builds until its run time would be comparable to at least JRuby.

@eregon Thank you for looking into it. Feel free to propose improvement that might also benefit other implementations too.

petergoldstein:
@pointlessone This splits out truffleruby into a scheduled (and also manual, upon request) workflow.  There's some duplication because we're reusing the push/pull_request CI config for the normal scheduled one.  So I simply added a truffleruby only config in another file.

spec/prawn_manual_spec.rb
```diff
@@ -4,12 +4,11 @@
...
-  when 'ruby'
-    '2c0279e0bff2a9120494a52aa46216c1871902b5e66a3537bd4d3cbd66db0096b43b6e1ae0e4e189b561c4db9fa1afacb6c41f260e3aaf942faae2fee352d35b'
-  when 'jruby'
-    '51baf6440907e9e38d22f50deafa91572aec1174e621c044ae077cfe3d4361982a505dae5f013dd06f64f38cb9b3a38d5a3f8f0903849591774e298a3c91d39a'
-  end
+  {
+    'ruby' => '2c0279e0bff2a9120494a52aa46216c1871902b5e66a3537bd4d3cbd66db0096b43b6e1ae0e4e189b561c4db9fa1afacb6c41f260e3aaf942faae2fee352d35b',
+    'jruby' => '51baf6440907e9e38d22f50deafa91572aec1174e621c044ae077cfe3d4361982a505dae5f013dd06f64f38cb9b3a38d5a3f8f0903849591774e298a3c91d39a',
+    'truffleruby' => '9cab8cd0736f239058ce8dc4b2915fe014e1461d434de456d9fbf140c2b69553f39accbe9acb38b974efd1e7d3e5dfddf77e99e0bdba56f43a8746a0e5fcf37b'
+  }.freeze
```
pointlessone, line 11: I wander why you changed it to a hash

.github/workflows/experimental.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ~/gems
+          key: gems-${{ matrix.ruby }}-${{ hashFiles('*.gemspec', 'Gemfile') }}-${{ github.sha }}
+          restore-keys: |
+            gems-${{ matrix.ruby }}-${{ hashFiles('*.gemspec', 'Gemfile') }}-
+            gems-${{ matrix.ruby }}-
+      - name: Install dependencies
+        run: |
+          gem install bundler
```
eregon, line 32: This is redundant, `setup-ruby` already install Bundler

.github/workflows/experimental.yml
```diff
@@ -0,0 +1,36 @@
...
+        ruby:
+          - truffleruby-head
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Gems cache
+        uses: actions/cache@v2
```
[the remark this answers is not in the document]
eregon: Ah to force the path to be `~/gems` and not `vendor/bundle` maybe?

eregon (commented):
Looks good to me

eregon:
I took another look at the performance issue (https://github.com/oracle/truffleruby/issues/2599) and I think I've found the culprit now. I'll try to fix it soon.

@petergoldstein Could you rebase this PR and address https://github.com/prawnpdf/prawn/pull/1246#discussion_r812994292 ?

The test suite runs in 1min24s for me locally so I feel it's more than fast enough for a scheduled CI job.
So maybe we can merge this PR?

.github/workflows/experimental.yml
petergoldstein: I'm aware of that, but there was a desire by the maintainer to be consistent with ci.yml. And as I recall a previous suggestion for removing that line from ci.yml was not accepted.

eregon: Oh OK then fair enough

eregon:
FWIW, the performance issue with `Prawn::Images::PNG#split_image_data` has been fixed in https://github.com/truffleruby/truffleruby/issues/2599

PR #1253, opened 2022-02-14
Clean up document state before raising Prawn::Errors::CannotFit

### LINKED ISSUE #1073: character_spacing bug after raising Prawn::Errors::CannotFit

petergoldstein:
Fixes #1073 

This is a rebased and rubocop-ed version of #1074   Thanks to @ndbroadbent for the original contribution.

I want to track down exactly what state is being cached (and not cleared) before this is merged, so I'm going to leave it as a draft PR for now.

### DIFF
--- a/lib/prawn/text/formatted/box.rb
+++ b/lib/prawn/text/formatted/box.rb
@@ -234,6 +234,9 @@ module Prawn
             end
           end
 
+          # Only raise CannotFit error after cleaning up document state
+          raise Errors::CannotFit if @line_wrap.cannot_fit
+
           unprinted_text.map do |e|
             e.merge(text: @document.font.to_utf8(e[:text]))
           end
@@ -548,18 +551,13 @@ module Prawn
         # size is reached
         def shrink_to_fit(text)
           loop do
-            if @disable_wrap_by_char && @font_size > @min_font_size
-              begin
-                wrap(text)
-              rescue Errors::CannotFit
-                # Ignore errors while we can still attempt smaller
-                # font sizes.
-              end
-            else
-              wrap(text)
-            end
+            wrap(text)
 
-            break if @everything_printed || @font_size <= @min_font_size
+            unless @disable_wrap_by_char && @font_size > @min_font_size
+              break if @everything_printed ||
+                @line_wrap.cannot_fit ||
+                @font_size <= @min_font_size
+            end
 
             @font_size = [@font_size - 0.5, @min_font_size].max
             @document.font_size = @font_size
--- a/lib/prawn/text/formatted/box.rb
+++ b/lib/prawn/text/formatted/box.rb
@@ -553,11 +553,8 @@ module Prawn
           loop do
             wrap(text)
 
-            unless @disable_wrap_by_char && @font_size > @min_font_size
-              break if @everything_printed ||
-                @line_wrap.cannot_fit ||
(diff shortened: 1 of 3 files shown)


petergoldstein:
@gettalong if you have any thoughts on the above, would be appreciated.  Thanks.

petergoldstein:
I've confirmed the above.  Updating pdf-core with `ensure` produces the correct behavior.  I'm going to put up a PR in `pdf-core` to resolve the issue there.

petergoldstein:
`pdf-core` PR is here - https://github.com/prawnpdf/pdf-core/pull/53

pointlessone:
@petergoldstein This is not up for merge any more, right?

PR #1256, opened 2022-02-18
Really fix encryption issue

gettalong:
The previous fix incorrectly escaped "\r" as "\\\r". However, it needs
to be escaped as "\\r" (so backslash followed by letter r).

This pull request needs https://github.com/prawnpdf/pdf-core/pull/55 to work since there the needed constant is defined.

### DIFF
--- a/lib/prawn/security.rb
+++ b/lib/prawn/security.rb
@@ -229,14 +229,14 @@ module PDF
         obj =
           ByteString.new(
             Prawn::Document::Security.encrypt_string(obj, key, id, gen)
-          ).gsub(/[\\\r()]/) { |m| "\\#{m}" }
+          ).gsub(/[\\\r()]/, STRING_ESCAPE_MAP)
         "(#{obj})"
       when Time
         obj = "#{obj.strftime('D:%Y%m%d%H%M%S%z').chop.chop}'00'"
         obj =
           ByteString.new(
             Prawn::Document::Security.encrypt_string(obj, key, id, gen)
-          ).gsub(/[\\\r()]/) { |m| "\\#{m}" }
+          ).gsub(/[\\\r()]/, STRING_ESCAPE_MAP)
         "(#{obj})"
       when String
         pdf_object(
--- a/spec/prawn/document/security_spec.rb
+++ b/spec/prawn/document/security_spec.rb
@@ -134,7 +134,7 @@ describe Prawn::Document::Security do
         )
       ).to eq(
         bin_string(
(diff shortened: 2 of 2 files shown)


petergoldstein:
@pointlessone @gettalong So this seems like a good example of the additional complexity introduced by a separate `pdf-core` gem.  We're going to have to do some gymnastics here to get CI to pass.

Rather than doing a `pdf-core` gem release and then upping the requirement for `prawn` (which would be one option), I suggest we doing something similar to what I did here - https://github.com/prawnpdf/prawn-table/blob/38b5bdb5dd95237646675c968091706f57a7a641/lib/prawn/table.rb#L32

Specifically we'd introduce a line like:

```
    STRING_ESCAPE_MAP = { '(' => '\(', ')' => '\)', '\\' => '\\\\', "\r" => '\r' }.freeze unless defined?(STRING_ESCAPE_MAP)
```

into `lib/prawn/security.rb`.

This is undoubtedly a bit of a hack, but it will get this PR unblocked and can be removed once we release a `pdf-core` version.

Thoughts?

gettalong:
@petergoldstein @pointlessone Wouldn't it be more "correct" to use the master branch of each repo in the CI pipeline instead of the released versions?

pointlessone:
Yeah, it probably would.

gettalong:
@petergoldstein I understand what you are saying but I think the exception you mentioned comes into play here since pdf-core and prawn are interdependent, as can be seen by this pull-request.

So even if we work around it and someone would use prawn from master seeing it as green, the whole system prawn+pdf-core wouldn't correctly work since the correct pdf-core wasn't yet released. This is the reverse from the case you have described.

In essence a user has to know that only released gems are safe to use. This should *always* be the case and if they uses some not-released code, they are own their own. Just my 2c.

petergoldstein:
Oh, and one additional benefit of that merge is that it'll make it easier to land updated encryption, PDF/A-lb, and likely other items in the future.

pointlessone:
I'd rather not merge the two gems. Not in a minor release anyway. And bumping the major just for the merge doesn't feel like an enough of a change from the user perspective. So if you have half a day maybe spend a quarter investigating what's left to decouple pdf-core from Prawn. It might turn out that there's not that much work left.

As for testing. Given that pdf-core and prawn are coupled right now I don't see an issue in testing against pdf-core's master. 

Customary we release both prawn and pdf-core at the same time. We can release pdf-core a day earlier and test and merge this PR right before the release of prawn.

pointlessone:
This was merged in git. Thank you.

PR #1268, opened 2022-06-08
fix: allow to use text and text_box in sequence with inline format flag

majioa:
- [x] !  allow to use text and text_box in sequence with inline format flag

### DIFF
--- a/lib/prawn/text.rb
+++ b/lib/prawn/text.rb
@@ -159,7 +159,7 @@ module Prawn
       if p
         p = [] unless p.is_a?(Array)
         options.delete(:inline_format)
-        array = text_formatter.format(string, *p)
+        array = string.is_a?(Array) ? string : text_formatter.format(string, *p)
       else
         array = [{ text: string }]
       end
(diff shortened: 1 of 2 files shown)


gettalong:
I'm sorry but I don't understand the reason of the change. Could you be a bit more expressive with respect to explaining the reason for the PR?

majioa:
Fox example I have a chain to draw text (as in the docs) but using the inline formatting. Since the excess_text after the call will contain an array of formatted text, so to keep the call stylistics to code with no inline formatting, and avoiding the exception, I just added the following two lines.

         excess_text = pdf.text_box(
            rendered(desc),
            width: 242,
            height: 242,
            overflow: :truncate,                                                   
            at: [0, y_position],                                                   
            inline_format: true                                                    
         )
         pdf.bounding_box([250, y_position], width: 200, height: 240) do           
            pdf.image(URI.open(icon_link.url), fit: [200, 240])                    
         end                                                                       
         pdf.text(excess_text, width: 450, inline_format: true)

PR #1278, opened 2023-01-01
Update CI workflow

mishina2228:
- Bump actions/checkout from v1 to v3
- Enable `bundler-cache` to cache gem dependencies
- Add Ruby 3.2 to the CI matrix
- Fix the RuboCop error in GitHub Actions
  ```
  Unable to find gem panolint; is the gem installed? Gem::MissingSpecError
  ```
  refs: https://github.com/rubocop/rubocop/issues/9832

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,24 +10,12 @@ jobs:
   code-style:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: 3.1.0
-      - name: Gems cache
-        uses: actions/cache@v2
-        with:
-          path: ~/gems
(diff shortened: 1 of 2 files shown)


gettalong:
If I remember correctly, there was a reason why the bundler cache was not used. Should be a comment on a pull request similar to this one.

pointlessone:
By default these actions install gems _inside_ the project root which messes up RuboCop check because it check all ruby files in the project root.  At least that's how I remember it. Maybe things have changed since then or I might've dome something wrong. In any case CI is clearly misconfigured…

mishina2228:
> By default these actions install gems _inside_ the project root which messes up RuboCop check because it check all ruby files in the project root. 

Yeah, that's why I had to change `.rubocop.yml` to make RuboCop ignore the ruby files in gem dependencies.
I've edited the description to mention it.

technicalpickles:
>  RuboCop check because it check all ruby files in the project root

Looks like this might be an upstream issue, found https://github.com/rubocop/rubocop/issues/9325 which I think describes it. Another fix for this seems to be this in .rubocop.yml:

```yaml
inherit_mode:
  merge:
    - Exclude
```

PR #1293, opened 2023-02-16
fix float precision bug in text box rendering

### LINKED ISSUE #1292: text_box with shrink_to_fit sometimes crops end of sentence
In rare cases, `Text#text_box` (https://www.rubydoc.info/gems/prawn/Prawn/Text#text_box-instance_method) with the `overflow: :shrink_to_fit` option causes part of the text to be cropped out. This is unexpected because if the text doesn't fix the box, it should be shrunk until it does, not cropped. 

This occurs specifically when the width of the box is _just_ big enough to fit the text and two float precisions errors occur. Say, for example, we want to put `"Two words"` in a text box of width 50.0. First, when `shrink_to_fit` runs to determine the font size, the width of "Two words" is computed as 49.999999.`shrink_to_fit` says, cool, this will fit, no need to reduce the font size or wrap the line. Then the line is actually printed and this time the width is computed as 50.00000001. This causes the line to be split and "words" is put on the second line, but if there isn't enough height, it will be cropped.

I have a commit with a spec that manually introduces these precision errors and has a failing spec here: https://github.com/unterkoefler/prawn/commit/aca004479435e74664ad348d87253dce96508d01

Happy to put up a pr to fix :)

unterkoefler:
fixes #1292


Jasmeet107 (approved):

lib/prawn/text/formatted/line_wrap.rb
```diff
@@ -101,7 +101,7 @@ def add_fragment_to_line(fragment)
                   @document.width_of(segment, kerning: @kerning)
                 end
 
-              if @accumulated_width + segment_width <= @width
+              if @accumulated_width + segment_width <= @width + 0.00001
```
pointlessone, line 104: Could you please add a test that demonstrates validity of this fix?

Also could you please explain the value of that number? Why can't it be smaller/bigger?

unterkoefler: [This commit](https://github.com/unterkoefler/prawn/commit/aca004479435e74664ad348d87253dce96508d01) adds a test that fails, but it does so by manually inserting the necessary rounding errors, so I didn't add it to this pr. Because this bug is so rare, I don't know a good way to add a test for it without changing the code.

unterkoefler: I chose the value `0.00001` arbitrarily. Seems like it's big enough fix the bug and small enough to not matter visually. The [ruby docs for float](https://ruby-doc.org/3.2.1/Float.html) link to [this page](https://github.com/rdp/ruby_tutorials_core/wiki/Ruby-Talk-FAQ#floats_imprecise) which uses `0.0001` as a delta. Happy to update to that instead if you think that would be better

gettalong: FYI In the geom2d gem which heavily uses floating point calculations I have implemented special comparison methods, see https://github.com/gettalong/geom2d/blob/master/lib/geom2d/utils.rb#L24-L34. The precision to use is arbitrarily set.

For Prawn I would probably choose something that reflects the output precision used in https://github.com/prawnpdf/pdf-core/blob/master/lib/pdf/core/pdf_object.rb#L16

unterkoefler: Thanks @gettalong! 

@pointlessone I updated to a value based on the precision and also updated one other place doing a similar comparison to use the same value

unterkoefler:
> I still don’t quite understand the root cause of the issue.

When we render a text box with shrink_to_fit enabled, we call `wrap` twice. The first time is to check if the text will fit in the box. If it doesn’t fit, the font size is reduced and we try again. When that’s done, the text should fit in the box. Then we call `wrap` again to actually draw the text.

In `wrap`, the check to see if the text will fit in one line without wrapping to the next is this:
```if @accumulated_width + segment_width <= @width```

The bug occurs specifically when the LHS `@accumulated_width + segment_width` (the entire width of the line) is equal to `@width` (the width of the text box). The first time through when we’re trying to figure out if the font is small enough, the float math works out such that `width of the line <= width of the text box`. Everything stays on the same line and `shrink_to_fit` says that it will fit at the current font size.

Now, `wrap` runs again, this time to do the actual rendering. This time, because float math is non-deterministic, `@accumulated_width + segment_width` is not `<= @width`. It’s very very slightly bigger. So now we think that the text won’t fit on one line, so the last word gets wrapped onto the next line. If the text box is too short, the second line doesn’t get rendered.

The assumption being violated here is that subsequent calls to `wrap` will produce text laid out the same way.

Also, regardless, `@accumulated_width + segment_width <= @width` is a problem because it’s comparing floats for equality without a delta.

PR #1294, opened 2023-02-22
POC: pdf-core Integration

gettalong:
In https://github.com/prawnpdf/prawn/pull/1256 we were talking about merging prawn and pdf-core.

Originally, I thought that it would only make sense if we remove the `pdf/core/*` namespace and integrate everything into the `prawn/*` namespace. However, I think the first step would just be moving all `lib/pdf` and `spec/pdf` files into the prawn repository which this pull request does.

By integrating pdf-core into prawn the release management gets easier and PR interdependencies like with #1256 won't happen. Since pdf-core isn't really used by anybody as a standalone library, I think going this way is okay.

Once this PR is merged, I think that several other PRs can be merged, too. There is at least #1256.


pointlessone:
I don't think we'll go with merging pdf-core into prawn for the next release for sure. And later if we decide to merge it'd probably be better to go straight to move everything into `Prawn` namespace. After all, that's how it was before the split.

gettalong:
Fair call.

I still think moving pdf-core into prawn as-is would make maintenance and release management easier. Since it is an internal dependency and nothing would change API wise, it would even be possible for the next release.

It may also open up possibilities for the community to contribute since I think once everything is in one repo and fully tested, it is easier to rewrite parts of pdf-core to use the `Prawn` namespace.

PR #1300, opened 2023-03-25
Add Docker with testing docker-compose config

mkrajewski90:
Simple Docker config to faster run tests without care of setup dev environment.

Reject if docker config is out of gem vision.


pointlessone:
Thank you for your contribution but this is not something I'm interested in supporting.

If you must use Docker to run tests, I suggest looking into [act](https://github.com/nektos/act).

PR #1301, opened 2023-03-25
Better background management

mkrajewski90:
I have a project with more complex background management than current basic setup.

What I need is:
- possible to set background for portrait pages and landscape pages mixed in one file
- disable background for some pages (like intro page or additional resources page)

I implemented Background class to better manage background cases. It's backward compatible and flexible to improve in the future. 

Tested in real, biz environment :)


pointlessone:
This feels like a too specific feature. Can this be implemented with whatever's currently available?

johnnyshields:
Portrait/landscape images seems too specific... the rest of it looks reasonable to add.

mkrajewski90:
> Portrait/landscape images seems too specific... the rest of it looks reasonable to add.

Prawn support only portrait and landscape layout, so wy too specific here?

Any ideas how you can see this?

pointlessone:
I feel like this can be relatively easily implemented with `on_page_create`. Probably, with less code, too.

mkrajewski90:
> I feel like this can be relatively easily implemented with `on_page_create`. Probably, with less code, too.

Thats exactly what we did before and I tell you it was really a nightmare. After implement this way it's abstract enough and easy to specify how the background should look like. 

If you feel that's too concrete you might close this PR, but I see after complex using in biz env that's really helpful.

johnnyshields:
> Prawn support only portrait and landscape layout, so wy too specific here?

Can't you just have "background_image" per page, and supply the right image?

mkrajewski90:
> > Prawn support only portrait and landscape layout, so wy too specific here?
> 
> 
> 
> Can't you just have "background_image" per page, and supply the right image?

No, let me give you an example - a pdf is with portrait layout content and one of the page is landscape layout table report that have too many data to be portrait layout. In the landscape it must be another image to the background.


Another option might be stay with basic Background class that have only background with image and scale, create and invoke them as I put in PR, than I can open it and add more specific logic (like additional params or enable/disable).

Issue #1303, opened 2023-04-28
Incompatibility with ruby 3.1


waghanza:
Hi @pointlessone,

I can see in https://github.com/prawnpdf/prawn/blob/8ceaa105f3e6038fac2eb4b84355d71b00d895f3/prawn.gemspec#L39 that `prawn` require **matrix**.

However, this line is not in `2.4.0`, and then this version could not be used with ruby **3.1**+.

May a new version be published ?

Regards,

PS : `matrix` has been removed => https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

gettalong:
See https://github.com/prawnpdf/prawn/issues/1273

PR #1304, opened 2023-04-28
Use default version of matrix

waghanza:
Hi,

In ruby < 3.1, we do not need this gems (it is bundled by default).

Regards,

---------

Related to https://github.com/prawnpdf/prawn/issues/1303

### DIFF
--- a/prawn.gemspec
+++ b/prawn.gemspec
@@ -36,7 +36,7 @@ Gem::Specification.new do |spec|
   ]
   spec.licenses = %w[PRAWN GPL-2.0 GPL-3.0]
 
-  spec.add_dependency('matrix', '~> 0.4')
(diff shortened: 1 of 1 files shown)


gettalong:
This doesn't make sense since the released gem will have a list of static dependencies. Your change would compare the version of Ruby that the person releasing the gem is using and selecting the matrix dependency based on this information. This would change nothing for the end user.

waghanza:
You say that `matrix` will be installed depending on which ruby version has the releaser @gettalong ? It is not my intention, and sorry for that. 

The biggest problem here is that `prawn` is not usable (at least the last version) with ruby >3.1.

pointlessone:
@waghanza Try master (or the current HEAD). Otherwise, it's a known issue. See #1195 .

waghanza:
I've subscribe to this issue :-)

PR #1320, opened 2023-11-10
Update line_width.rb (in manual)

PhilippePerret:
For consistency (use of function rather than defining variable)


pointlessone:
Thank you for your contribution but I'm not sure what is the purpose of this change.

Feel free to reopen with a clarification.

pointlessone:
Right, I see now. This manual section doesn't mention the regular DSL-style API, though. It only points out the distinction between attribute assignment and local variable assignment as it's a common gotcha.

PhilippePerret:
> This manual section doesn't mention the regular DSL-style API, though. It only points out the distinction between attribute assignment and local variable assignment as it's a common gotcha.

I agree. 

But the manual is a whole, isn't it? And I find that it creates confusion to always have the use of one method throughout the manual, and then suddenly switch to using the explicit definition of a variable... The subject of the section is not "you can also define attribute by assignment", it's "the line width".
I had to look into the code to understand why the logic was changing all of a sudden.
It's not consistent.

But again, if you don't find it relevant, no problem.

Have fun!

PR #1323, opened 2023-12-13
Add options :position and :vposition when embedding image

mirelon:
Possible values:
```rb
position: :center
position: :right
vposition: :center
vposition: :bottom
```


pointlessone:

Could you provide a bit more context please? Is there a way to achieve the same effect currently? Why does this have to be in Prawn?

Issue #1307, opened 2023-08-07
Error in the manual? (page 11)


bgvo:
In page 11 is said:

```
... By defining your own document method, as shown in the example, you will be able...
```

But in the example, such a method (`document`) is not defined. 

<img width="784" alt="image" src="https://github.com/prawnpdf/prawn/assets/3684835/83beba77-b24f-4e0a-999c-d6103157df69">

pointlessone:
You're quite correct. This is a bug.

You can find an example of the method in #1258.

pointlessone:
This has been [fixed](https://github.com/prawnpdf/prawn/blob/1d4469d5b6df09fe417f9adb60070cee2b5195c2/manual/basic_concepts/view.rb#L40-L42). Thank you for bringing my attention to it.

PR #1324, opened 2024-01-16
add example of a decument method to the manual

Juraj-Sulimanovic:
[Issue 1307](https://github.com/prawnpdf/prawn/issues/1307)

The example that is mentioned in the manual is missing. 
I just added the example mentioned on the issue itself.


pointlessone:

However, I don't think this is a very good addition. Manual describes how the method can be defined in order to get a custom document and here a default one is defined instead. I believe the manual is clear enough on the optionality of the method and its purpose. So if you want to demonstrate that then at least add some options to the document so that it was clear why the method is present in the example.

Juraj-Sulimanovic:
> Thank you for your contribution.
> 
> However, I don't think this is a very good addition. Manual describes how the method can be defined in order to get a custom document and here a default one is defined instead. I believe the manual is clear enough on the optionality of the method and its purpose. So if you want to demonstrate that then at least add some options to the document so that it was clear why the method is present in the example.

Thank you for the constructive comment. I added some options in the method.
The failing spec is related to a different commit.

pointlessone:
Thank you for your contribution. This has been merged outside of GitHub.

PR #1327, opened 2024-01-22
Fix missing glyph widths

pointlessone:
This is an attempt to address an issue raised in https://github.com/prawnpdf/prawn/pull/1322#issuecomment-1895462228


mojavelinux:
This fixes the original issue that I raised in https://github.com/prawnpdf/prawn/pull/1322#issuecomment-1895462228. However, I'm detecting one other failure in the Asciidoctor PDF test suite that is coming from changes in Prawn. I don't yet understand the failure. I'll report back once I do.

mojavelinux:
Ha! The test is failing because Prawn is fixed. It was previously truncating Unicode characters outside of the 4-digit range (e.g., U+000f089a would become U+f089). Now the code does what it's supposed to do (preserves the original char code) and I just need to update the assumption in the test. False alarm.

With that out of the way, I can confirm that this fix addressed the compatibility issues.

pointlessone:
Good to hear. Thank you. I will land this soon.

Issue #1343, opened 2024-03-12
Grid broken if no gutter defined


bvogel:
The following code block in grid.rb will break if no gutter is defined:
```
      def apply_gutter(options)
        if options.key?(:gutter)
          @gutter = Float(options[:gutter])
          @row_gutter = @gutter
          @column_gutter = @gutter
        else
          @row_gutter = Float(options[:row_gutter])
          @column_gutter = Float(options[:column_gutter])
          @gutter = 0
        end
      end
```
the commit https://github.com/prawnpdf/prawn/commit/be66ddd0a804a03f6d7979bc773ea0aad0e7ec29
replaced `options[:row_gutter].to_f` which happily returned `0.0` for `nil` values with `Float(options[:row_gutter])` which will break with `TypeError: can't convert nil into Float`.  I'd have considered this breaking change that would have merited a mayor version change.

afdev82:
I have hit this bug too...
What should be the fix here? Rechange `Float()` with `to_f` or check for `nil`?
I could submit a PR

datanoise:
We ran into this issue too. It would be nice if it is fixed going forward. It is definitely a breaking change.

pointlessone:
@datanoise I second that. It'd be absolutely marvellous if the bug got fixed.

andreavocado:
There is a workaround for this with same behaviour. Define the `gutter` argument with zero:

before:
`define_grid(rows: 1, columns: 2)`

after:
`define_grid(rows: 1, columns: 2, gutter: 0)`

afdev82:
I have already submitted a PR for that.

> @datanoise I second that. It'd be absolutely marvellous if the bug got fixed.

VoidF0rk:
Have hit the same problem. Considering a PR exists to fix this, and looks reasonable, please address; options being mandatory breaks convention.

PR #1345, opened 2024-03-19
Fix for issue #1343

afdev82:
Fallback to 0.0 in case of no gutter

### DIFF
--- a/lib/prawn/grid.rb
+++ b/lib/prawn/grid.rb
@@ -145,12 +145,12 @@ module Prawn
 
       def apply_gutter(options)
         if options.key?(:gutter)
-          @gutter = Float(options[:gutter])
+          @gutter = Float(options[:gutter] || 0.0)
           @row_gutter = @gutter
           @column_gutter = @gutter
         else
-          @row_gutter = Float(options[:row_gutter])
-          @column_gutter = Float(options[:column_gutter])
+          @row_gutter = Float(options[:row_gutter] || 0.0)
+          @column_gutter = Float(options[:column_gutter] || 0.0)
           @gutter = 0
         end
       end


lib/prawn/grid.rb
```diff
@@ -145,12 +145,12 @@ def subdivide(total, num, gutter)
 
       def apply_gutter(options)
         if options.key?(:gutter)
-          @gutter = Float(options[:gutter])
+          @gutter = Float(options[:gutter] || 0.0)
```
andreavocado, line 148:
```suggestion
          @gutter = Float(options.fetch(:gutter, 0.0))
```

lib/prawn/grid.rb
```diff
@@ -145,12 +145,12 @@ def subdivide(total, num, gutter)
...
           @row_gutter = @gutter
           @column_gutter = @gutter
         else
-          @row_gutter = Float(options[:row_gutter])
-          @column_gutter = Float(options[:column_gutter])
+          @row_gutter = Float(options[:row_gutter] || 0.0)
```
andreavocado, line 152:
```suggestion
          @row_gutter = Float(options.fetch(:row_gutter, 0.0))
```

lib/prawn/grid.rb
```diff
@@ -145,12 +145,12 @@ def subdivide(total, num, gutter)
...
+          @column_gutter = Float(options[:column_gutter] || 0.0)
```
andreavocado, line 153:
```suggestion
          @column_gutter = Float(options.fetch(:column_gutter, 0.0))
```

andreavocado (commented):
There are specs missing when `options` is empty or `options[:*gutter]` is nil.

lib/prawn/grid.rb
afdev82: Even better in this way, but I have found already this style in another part of the code, that's why I have used that.
We should change maybe also [there](https://github.com/prawnpdf/prawn/blob/ce374035905ef7691ddfa2dc015f4ea734d1f737/lib/prawn/document/bounding_box.rb#L534)

afdev82:
Yes, then I will add the specs.

lib/prawn/grid.rb
pointlessone: I appreciate deference to existing code style but given that the project is relatively old I'd not pay very much attention to it.

I like `fetch`.

afdev82:
I have added a simple test case.

PR #1350, opened 2024-06-02
Fix typo in image scaling section of the manual

LukasSkywalker:


.tool-versions
```diff
@@ -0,0 +1 @@
+ruby jruby-9.4.7.0
```
pointlessone, line 1: This doesn't seem like a relevant change.

pointlessone:

PR #1363, opened 2024-09-27
Fixes circular require warning

kbrock:
fixes:

```
warning: lib/prawn/fonts.rb:9: warning: loading in progress, circular require considered harmful - lib/prawn/font.rb
```

### DIFF
--- a/lib/prawn/fonts.rb
+++ b/lib/prawn/fonts.rb
@@ -6,7 +6,6 @@ module Prawn
   end
 end
 
-require_relative 'font'
 require_relative 'fonts/afm'
 require_relative 'fonts/ttf'
 require_relative 'fonts/dfont'


radwo:
Hey team!

Any idea if this is going to be merge?

kbrock:
update:
- rebased to ensure an easy merge.

Thanks for the ping @radwo
I too hope this can be merged

practicingdev:
I was curious about why this was happening to begin with and it looks like it's due to the compatibility shim that sets constants on `Font::*` to point to the `Fonts::**` namespace.

`Fonts::AFM` and `Fonts::TTF` do inherit from `Font` so it does need to be loaded first. However, because this happens in `font.rb` and the `fonts.rb` file is *not* loaded from anywhere else, it does seem safe to simply remove that line.

Did not attempt to run the code, just dropping a note based on what I found reading things.

kbrock:
This failure is unrelated.
It only fails on ruby head and no other version.
The failure is based upon a change in encoding.

PR #1364, opened 2024-10-22
feat: Pixels to Points

wilsondealmeida:
### Motivation
Convert pixels to points.
Useful when you are working with images.

Example:
```ruby
1.px
=> 0.75 # points

800.px
=> 600.0 # points
```

### DIFF
--- a/lib/prawn/measurement_extensions.rb
+++ b/lib/prawn/measurement_extensions.rb
@@ -79,4 +79,11 @@ class Numeric
   def pt
     pt2pt(self)
   end
+
+  # Convert from pixels to points.
+  #
+  # @return [Number]
+  def px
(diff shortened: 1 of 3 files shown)


pointlessone:

PDF has no concept of a sized pixel as such. So length can not be converted to pixels and vice versa. This conversion is completely a CSS thing (CSS Values and Units Module Level 3, [Absolute Lengths](https://www.w3.org/TR/css-values-3/#absolute-lengths)). I understand when you might want this but this is outside of PDF so I'd rather keep it outside of Prawn.

I suggest you use a refinement in your application.

---

I'm closing this for the reason above but feel free to continue discussion and suggest other improvements.

Issue #1346, opened 2024-03-25
The Return of the Embedded Font Issue


TastyPi:
Prawn 2.5.0 appears to have reintroduced this old "Cannot extract the embedded font" issue https://github.com/prawnpdf/ttfunk/issues/73

Using the exact same code to render a PDF using 2.4.0 does not have this issue.

pointlessone:
Do you experience this issue if you disable font subsetting?

TastyPi:
I tried with `subset: true` and `subset: false`

pointlessone:
Could you please provide a minimal reproduction script and fonts you have issues with?

pointlessone:
Also could you please list versions of the gems you use?

```sh
gem list --all --exact prawn ttfunk pdf-core
```

TastyPi:
I tried a minimal example that just sets the font, but that did not reproduce the issue so it's definitely something deeper. Our PDF generation code is pretty complex, so it'll take me a while to create a minimal example, not sure when I'll find the time :confused:.

Before downgrading (with issue):

```
bundle list | grep -E 'prawn|ttfunk|pdf-core'
  * pdf-core (0.10.0)
  * prawn (2.5.0)
  * ttfunk (1.8.0)
```

After downgrading (no issue):
```
bundle list | grep -E 'prawn|ttfunk|pdf-core'
  * pdf-core (0.9.0)
  * prawn (2.4.0)
  * ttfunk (1.7.0)
```

TastyPi:
Thanks to the magic of Ruby, I've managed to log every call made to the `Prawn::Document` instance, and successfully reproduced the issue. I'm just cleaning it up to remove specific details and make it minimal.

pointlessone:
Thanks.

coorasse:
Thanks for the deep dive @TastyPi, you helped us! 
Will you open an issue on ttfunk, or is the team aware of it?
We locked ttfunk `gem 'ttfunk', '< 1.8.0'` in the meantime.

TastyPi:
Filed https://github.com/prawnpdf/ttfunk/issues/102, I wasn't sure if @pointlessone was going to do it as they appear to be the main ttfunk developer.

pointlessone:
[...]

PR #1374, opened 2025-03-02
Downgrade ttfunk

dacook:
- #1346

The issue in ttfunk remains unsolved, meaning this version of Prawn is also broken, and has been for some time. This wastes time for each user of the gem, having to debug and research the the problem until they find some clues in issue comments and try those.

So I suggest that Prawn reverts to the previous version to make it work out of the box. Is it this simple?

### DIFF
--- a/prawn.gemspec
+++ b/prawn.gemspec
@@ -48,8 +48,8 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_dependency('matrix', '~> 0.4')
-  spec.add_dependency('pdf-core', '~> 0.10.0')
-  spec.add_dependency('ttfunk', '~> 1.8')
+  spec.add_dependency('pdf-core', '~> 0.9.0')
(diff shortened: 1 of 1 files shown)


dacook:
The CI failure may be due to the gem cache. I was able to `bundle install` fine, but it seems that CI is relying solely on cached gems. 

Co-Pilot makes a good point:
> The failing job in the CI workflow appears to be related to the cache action for the gems. The cache hit indicates that it successfully restored the cache but the subsequent steps might be failing.

samyouel:
Hey! Any news on this PR? It would be amazing and would make our lives so much easier again if that worked out of the box again. 
A bit strange, why https://github.com/prawnpdf/prawn/issues/1346 continues to be closed, although the issue is still relevant?

dacook:
@pointlessone can you please review and merge if you are ok with this?

Of course, it would be better to resolve the underlying issue, but that has been unresolved for a long time, so I believe this is worth doing to resolve the issue which is affecting a lot of people.

PR #1377, opened 2025-04-29
Revise confusing 'valign' error message for text boxes

lasaldan:
Problem:
Using an unsupported symbol for the value of 'valign' on a text box throws a misleading error message. 

Solution:
This adjusts the error message to match the actual values supported by the code.

### DIFF
--- a/lib/prawn/text/formatted/box.rb
+++ b/lib/prawn/text/formatted/box.rb
@@ -589,7 +589,7 @@ module Prawn
             @at[1] -= (@height - height)
           else
             raise ArgumentError,
-              'valign must be one of :left, :right or :center symbols'
+              'valign must be one of :top, :bottom or :center symbols'
           end
 
           @height = height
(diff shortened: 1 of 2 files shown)


pointlessone (approved):

pointlessone:
Would you please fix the specs?

lasaldan:
@pointlessone Sorry for not updating the specs the first time around. They are passing now, other than the PrawnByExampleManual hash tests. Looks like those broke around the bump of prawn-dev to 0.5 back in January. 

I did try to investigate those failing tests, but couldn't determine definitively what causes the hash mismatch. I pulled the prawn-manual_builder gem so I could build the PDF manual locally, but the hash of the local PDF was indeed different than what was expected in the test. Not sure if the test needs the hash updated to the new PDF hash, or if there is something in the generated PDF that is incorrect causing the hashes to be different. 

Either way, that's a different issue and PR. Let me know if this one needs anything else. Thanks so so much for Prawn.

pointlessone:
Sorry, I meant only these specs. The hash one is tracked elsewhere and is a bit more involved. I'll get to it eventually.

Thank you.

PR #1388, opened 2026-02-07
Updated prawn-dev to 0.6

benjaminysmall:
The prawn-dev gem updated its version a few days ago which appears to have broken several of the CI tests. This PR tries to fix them by matching the version update.


benjaminysmall:
Instead, it appears _all_ the tests are now broken! Whoops. Going to continue iterating.

pointlessone:
Yeah, I know. Manual builder has some conflicting deps. That needs to be fixed first. IIRC, it's Prism. It's not a simple version bump either, the API has changed. It's probably not a huge change but still.

benjaminysmall:
Makes sense. I think I'll just close this PR then. Thanks and sorry for the spam!

pointlessone:
No problem. I appreciate the initiative.

Issue #1295, opened 2023-02-24
Advanced text layout


pointlessone:
Right now (2.4.0 and all versions before) Prawn implements a very simple text layout. It basically renders individual code points. The code points are used as they are present in the input string. This works kinda OK for Latin-based alphabets and hieroglyphic alphabets that don't use any fancy font techniques but doesn't work at all for more complex scripts.

- #830: Hindi seem to actively use glyph substitutions based on the context
- #921: Arabic uses different glyphs depending on where the "letter" is (start, middle, or end of the word), requires glyph substitutions
- #1098: This is not exactly an issue with the advanced fonts features but it's an issue with our basic text segmentation algorithm and maybe how we measure widths of individual code points instead of composite "characters"
- #1126: another example of Thai requiring more advanced text layout
- #1125: This script also extensively uses glyph substitution

There are no issues for this but this basic font rendering also prevents use of ligatures or contextual alternatives provided by fonts.

---

This issue is a canonical explantation for the lack of support of certain scripts. Community is welcome to work on a more advanced text layout.

jyxjjj:
I didn't understand, For example, will JetBrians Mono no ligatures? or with ligatures?
There was no ligature settings, should i open a new issue?

pointlessone:
@jyxjjj I believe the font does have ligatures but Prawn doesn't support ligatures. You can open a new issue but I'm not planning to work on anything like it in the near future. I will gladly accept a contribution though.

johnnyshields:
I've added multi-line support for Arabic here, the code is a mess but it mostly works: https://github.com/cropio/prawn-rtl-support/pull/5

I investigated using [Harfbuzz](https://github.com/harfbuzz/harfbuzz) for some of the other scripts. It's not clear to me whether basic ligature support at the font-level will fix them, or if if we need to transform the glyphs at the string-level.

johnnyshields:
FYI someone already implemented Harfbuzz for an equivalent library in Python called fpdf2:
- Harfbuzz support: https://github.com/py-pdf/fpdf2/pull/820
- Bidi support: Process bidirectional text in conformance to Unicode standards  py-pdf/fpdf2#1096

@pointlessone would you accept a PR for Prawn that adds "text shaping" functionality to core, as fpdf2 has done?

szanni:
I would love to see this implemented. It seems like Harfbuzz is the industry standard. From browsers to XeTex. Tried and proven.

[...]

PR #1392, opened 2026-04-02
Add Arabic text shaping support

alghanim:
## Problem

Prawn does not perform OpenType text shaping (GSUB init/medi/fina/isol features). Arabic is a cursive script where each character has up to 4 positional forms (isolated, initial, medial, final) depending on its joining context. Without shaping, all Arabic characters render in their isolated form — disconnected and completely unreadable.

This is a long-standing issue affecting all Arabic, Farsi, and Urdu users of Prawn.

## Solution

Add `Prawn::Text::ArabicShaping` module that converts Arabic characters to their correct Unicode Presentation Forms (U+FE70-U+FEFF, U+FB50-U+FDFF) based on joining context. The shaping is integrated into `formatted_text` and `draw_text` so it works automatically.

### How it works

1. Scan text for Arabic character runs
2. For each run, separate base characters from diacritical marks
3. Apply mandatory Lam-Alef ligatures
4. Determine joining context (what connects to what)
5. Select the correct presentation form (isolated/initial/medial/final)
6. Reassemble with marks preserved

### Supports

- All standard Arabic letters (U+0621-U+064A)
- Extended Arabic: Farsi (Peh, Tcheh, Gaf, Farsi Yeh), Urdu (Tteh, etc.)
- Mandatory Lam-Alef ligatures (4 variants)
- Diacritical marks (tashkeel) preservation
- Tatweel (kashida) joining
- Zero performance impact on non-Arabic text (early `return` when no Arabic detected)

### Integration

- `formatted_text`: shapes each fragment's `:text` before rendering
- `draw_text`: shapes text before encoding normalization
- Works with Prawn's existing `direction: :rtl` for proper right-to-left layout

## Test plan

RSpec tests included covering:
- Basic shaping (initial, medial, final, isolated forms)
- Lam-Alef ligatures (isolated and final)
- Diacritical marks preservation
- Mixed Arabic/Latin text
- Extended Arabic characters (Farsi, Urdu)
- Right-joining characters (Alef, Dal, etc.)
- Edge cases (empty, nil, non-Arabic passthrough)


pointlessone:
@alghanim Was AI used for this?

alghanim:
Yes of course

pointlessone:
Are you familiar with the writing system and all the languages here? Did you read, understand and verify the code?

alghanim:
Hi, thanks for the review.

Yes, I'm a native Arabic speaker and this is for our organization's production deployment. I've verified the code extensively.

**Test results (25 RSpec examples, 0 failures):**

```
.........................
Finished in 0.00526 seconds
25 examples, 0 failures
```

Tests cover:
- All 4 positional forms (isolated, initial, medial, final)
- Right-joining characters (Alef, Dal, Waw, etc.)
- Dual-joining characters (Beh, Teh, Seen, etc.)
- All 4 Lam-Alef mandatory ligatures (plain, Madda, Hamza above, Hamza below)
- Diacritical marks (tashkeel) preservation
- Tatweel (kashida) joining
- Extended Arabic: Farsi Yeh (U+06CC), Peh (U+067E), Gaf (U+06AF)
- Non-UTF-8 encoding passthrough (Shift_JIS etc.)
- Mixed Arabic/Latin text
- Full sentence shaping verification
- Prawn document integration with TTF fonts

The mapping table follows the Unicode Arabic Shaping specification — each character maps to its presentation forms in the Arabic Presentation Forms-B block (U+FE70-U+FEFF) and Forms-A block (U+FB50-U+FDFF). The joining algorithm classifies characters as dual-joining, right-joining, or non-joining and selects the correct form based on neighboring characters' joining capabilities.

Note: The CI "edge" test failures are caused by `prawn-manual_builder`'s `prism` dependency failing to install — this is a pre-existing issue unrelated to this PR. The "release" tests and code-style check should pass (code-style already passed in the last run).

pointlessone:
Thank you.

I know about the manual builder. It will be fixed. Eventually.

Would you please recommend some reading material for someone who's new to Arabic scripts that would help with evaluation of these changes?

One specific question I have is… This seems to replace generic (isolated) code points with positional form code points. Does it have any effect on the meaning of text? Does it interact in unintended ways external features such as text extraction or, say, search? For example, if I search for a word that consists of generic code points will common software find it in a document generated with this code?

alghanim:
Good questions.

**Reading material:**
- [Unicode Standard, Chapter 9: Middle East-I (Arabic)](https://www.unicode.org/versions/Unicode15.0.0/ch09.pdf) — sections 9.2 (Arabic) and 9.2.1 (Cursive Joining) are the most relevant
- [UAX #9: Unicode Bidirectional Algorithm](https://unicode.org/reports/tr9/) — for understanding text direction
- [Arabic Presentation Forms-B](https://www.unicode.org/charts/PDF/UFE70.pdf) — the Unicode chart showing the exact presentation form codepoints used in this implementation

**On meaning and text extraction:**

The presentation forms (U+FE70-U+FEFF) are defined by Unicode as **canonical equivalents** of the base Arabic characters — they carry the same semantic meaning. The only difference is visual: they encode the positional glyph variant.

However, you raise a valid concern about searchability. PDF viewers like Adobe Acrobat and most modern PDF readers handle this correctly because they normalize Arabic presentation forms back to their base characters during text extraction and search. This is standard behavior defined in the PDF specification (ISO 32000) via the ToUnicode CMap.

That said, **some** simpler PDF tools might not normalize. This is a known tradeoff in the PDF world — the same tradeoff that exists in every PDF library that renders Arabic (e.g., ReportLab, iText, wkhtmltopdf all use presentation forms).

The alternative would be to implement a full OpenType shaping engine (like HarfBuzz) which uses GSUB tables to select the correct glyph IDs directly — keeping the base codepoints in the PDF content stream. But that's a significantly larger effort and would require deep integration with ttfunk's font parsing. The presentation forms approach is the standard solution used by most PDF libraries that don't have a built-in shaping engine.

**In summary:** No change in meaning. Major PDF readers search correctly. The tradeoff is well-understood and widely accepted in the PDF ecosystem.

gettalong:
> However, you raise a valid concern about searchability. PDF viewers like Adobe Acrobat and most modern PDF readers handle this correctly because they normalize Arabic presentation forms back to their base characters during text extraction and search. This is standard behavior defined in the PDF specification (ISO 32000) via the ToUnicode CMap.

What exactly do you mean by that with reference to the ToUnicode CMap? Is Adobe Acrobat normalizing the presentation forms or is it done by the ToUnicode CMap?

pointlessone:
@gettalong I read is the document has to provide `ToUnicode` object with the font that maps positional glyphs to isolated code points, like any proper fond should.

alghanim:
@gettalong Good clarification question.

Both mechanisms exist, and they work together:

1. **ToUnicode CMap (in the PDF):** When Prawn embeds a TTF font subset, ttfunk generates a ToUnicode CMap that maps glyph IDs to Unicode codepoints. If the font's cmap table maps presentation form codepoints (e.g. U+FEE3 MEEM INITIAL) to their glyphs, the ToUnicode CMap will map those glyph IDs back to the presentation form codepoints. This is what @pointlessone is referring to.

2. **PDF reader normalization:** Adobe Acrobat and other readers have built-in Unicode normalization that maps Arabic Presentation Forms (U+FE70-U+FEFF) back to their base characters (U+0621-U+064A) during text extraction and search. This is part of the reader's text extraction pipeline, not the PDF itself. The [Unicode Character Database](https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt) defines decomposition mappings for all presentation forms (e.g. `FEE3;ARABIC LETTER MEEM INITIAL FORM;Lo;0;AL;<initial> 0645`  — the `<initial> 0645` means it decomposes to base MEEM U+0645).

So the answer is: the ToUnicode CMap preserves the mapping at the glyph level, and the PDF reader does the semantic normalization from presentation forms back to base characters. Both are standard behavior.

gettalong:
@alghanim I'm aware of the PDF internals and how the ToUnicode CMap work. And I'm not sure if I'm speaking with an AI or yourself right now.

Anyway,  you wrote the sentence *"This is standard behavior defined in the PDF specification (ISO 32000) via the ToUnicode CMap."* as if it also refers to the mapping of presentation forms back to base characters. This could certainly be done by outputting the glyph of a presentation form but mapping it to the Unicode character of the base form but I'm not sure if you meant that (and the code doesn't seem to do that).

khaledhosny:
Note that using Arabic Presentation Forms is a hack for applications that can’t use a proper shaping engine, it is not a standard nor a recommended solution. It supports shaping a limited subset of Arabic script (any Arabic character added to Unicode after the initial batch lacks an encoded presentation form). It is also a very large limited form of shaping as it does not handle OpenType substitutions or positioning and works only (in a limited way) with very simple Arabic fonts (for example, nastaliq fonts preferred for Urdu text, like Noto Nastaliq Urdu, will not work with this way).

PDF extraction is also a concern, but presentation forms end up in the PDF stream rather often (regardless of the method of shaping) and most PDF readers should be prepared to handle it by now (though such normalization is not required by PDF spec AFAIK). Doing this property also requires a mix ot ToUnicode and ActualText, since not all forms of glyph to codepoint mappings that can happen with Arabic fonts are supported by ToUnicode alone.

johnnyshields:
I've already added Arabic support here: https://github.com/prawn-rtl-support/prawn-rtl-support/pull/5

See this comment: https://github.com/prawnpdf/prawn/issues/1295#issuecomment-2677695497 The right approach will be to implement Harfbuzz library, to cover text shaping in all languages.

